PR #6, opened 2024-07-31
Update Config for RichTextQuote and update solana-lib version to 2.39.0 

d-la:
- Switches the field type of `quote` from `text` to `richText` to give more control over the look of the text.
- Updates solana-lib to `2.39.0` to pull updates to the RichTextQuote component so it uses `HtmlParser` to render the rich text


catalinred (approved):
Lgtm, ty @d-la!

PR #8, opened 2024-08-02
Remove alias config causing all Nextjs App directory paths to error

ZYJLiu:
Alias results in nextjs app dir paths to throw `TypeError: Cannot read properties of null (reading 'useContext')`
CleanShot 2024-08-02 at 13 26 33@2x

Removing following code from next.config.js fixes issue

```ts
      // Alias configuration
      config.resolve.alias = {
        ...config.resolve.alias,
        react: path.resolve(__dirname, "node_modules/react"),
      };
```
CleanShot 2024-08-02 at 13 27 39@2x

Code was added in [this PR](https://github.com/solana-foundation/solana-com-legacy/pull/1337). Solana-com seems to run fine with it removed.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -113,12 +113,6 @@ const moduleExports = () => {
       );
       imageLoaderRule.exclude = /\.inline\.svg$/;
 
-      // Alias configuration
-      config.resolve.alias = {
-        ...config.resolve.alias,
-        react: path.resolve(__dirname, "node_modules/react"),
-      };
-
       return config;
     },
 
--- a/next.config.js
+++ b/next.config.js
@@ -6,7 +6,6 @@ const withBundleAnalyzer = require("@next/bundle-analyzer");
 const { i18n } = require("./next-i18next.config");
 const rewritesAndRedirectsJson = require("./rewrites-redirects.json");
 const { builder } = require("@builder.io/sdk");
-const path = require("path");
 
 const securityHeaders = [
   {


catalinred:
Oh, although the standalone solana-com runs just fine with the above removed

catalinred (commented):
The standalone solana-com runs just fine with the above alias config therefore the config was removed when initing this repo in the first place.

But noticed that if you remove the above and then you [link](https://github.com/solana-foundation/solana-lib?tab=readme-ov-file#linking) the [solana-lib](https://github.com/solana-foundation/solana-lib) for local Builder testing purposes, you can see some errors are thrown too. 

So I re-added in https://github.com/solana-foundation/solana-com/commit/7d6cea6c290cf31a5fafd1d24fb3790aa565dca7. 

Open to suggestions here, maybe this alias should be removed indeed and only temp added when someone is working with solana-lib locally? Or is there a better way to handle this?

ZYJLiu:
> Open to suggestions here, maybe this alias should be removed indeed and only temp added when someone is working with solana-lib locally? Or is there a better way to handle this?

Can it be a conditional config based on a `.env` variable?

catalinred:
> > Open to suggestions here, maybe this alias should be removed indeed and only temp added when someone is working with solana-lib locally? Or is there a better way to handle this?
> 
> Can it be a conditional config based on a `.env` variable?

I don't think we have an .env variable to be specific for linking projects and to be used in this case but let me know if you have something in mind .

On removing the alias config, the part where the [`path` is `require`d](https://github.com/solana-foundation/solana-com/blob/main/next.config.js#L9) should be also removed.

catalinred (approved):
Until we get to a better conditional, we'll remove this config for now.

PR #9, opened 2024-08-04
Github footer now directs to this website's git

Tamgros:
The footer button for github was directing to /solana-labs. I'm assuming that was a legacy from when labs owned the site and client. Now that this site is owned by a different repo, I think it should be pointed there so contributors know where to go.

There's also an argument for a github link to direct to a protocol client repo, but that should live somewhere else because that's 3 different repos at this point.

### DIFF
--- a/rewrites-redirects.json
+++ b/rewrites-redirects.json
@@ -9,7 +9,7 @@
   "redirects": [
     { "source": "/reddit", "destination": "https://reddit.com/r/solana" },
     { "source": "/telegram", "destination": "https://t.me/solana" },
-    { "source": "/github", "destination": "https://github.com/solana-labs" },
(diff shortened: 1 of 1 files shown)


catalinred (commented):
Thank you @Tamgros! I'm wondering if we should better direct to the parent account (https://github.com/solana-foundation/) instead? 

Also, can you please run Prettier? :)

Tamgros:
Ran prettier.

I'd vote on solana-com repo because it's clear what the footer relates to. Linking to solana-foundtion org level has so much extra stuff. The website itself tries to point people to the right repo, whether that be a protocol, SDK, docs, etc. 

I don't feel strongly, shipping anything is fine with me. As long as it's not labs's repo.

catalinred (approved):
Lgtm @Tamgros, thank you!

Issue #32, opened 2024-09-09
The Superminority value is static


catalinred:
The Superminority  (aka Nakamoto coefficient) value is currently hardcoded on https://solana.com/validators.

PR #33, opened 2024-09-09
Compute the superminority value

catalinred:
This resolves https://github.com/solana-foundation/solana-com/issues/32.

The Superminority value is now fetched and computed within `useTransactionStats`, and then displayed accordingly in `ValidatorsCards.js`.

Preview on [/validators](https://solana-com-git-feat-compute-superminority-solana-foundation.vercel.app/validators):

image


catalinred:
@tigarcia would you be able to give it a look and confirm this works as expected here?

src/hooks/useTransactionStats.js
```diff
@@ -1,68 +1,81 @@
...
+export const fetchSuperminority = async () => {
+  try {
+    const voteAccounts = await makeRPCCall({
+      method: "getVoteAccounts",
+      rpcNodeURL,
+    });
+
+    // Sort validators by stake in descending order
+    const sortedValidators = voteAccounts.result.current.sort((a, b) =>
+      Number(BigInt(b.activatedStake) - BigInt(a.activatedStake)),
```
tigarcia: Doesn't this have the potential to fail? For example, Number(very large int - very small int).  You could rewrite it to be something like:

const sortedValidators = voteAccounts.result.current.sort((a, b) => {
      const diff = BigInt(b.activatedStake) - BigInt(a.activatedStake);
      if (diff > BigInt(0)) {return 1;}
      else if (diff < BigInt(0) { return -1;}
      else { return 0; }
   }

src/hooks/useTransactionStats.js
```diff
@@ -1,68 +1,81 @@
...
  *
- * @param {boolean}   visible                   Only fire new queries when visible.
- * @param {number}    performanceUpdateSeconds  Delay before next query.
- * @param {number}    sampleHistoryHours        How many hours (60min.) the query should go back.
- * @param {boolean}   getLiveTransactionCount
- * @param {boolean}   getCurrentValidatorNodes
- * @returns {{availableStats: boolean, avgTps: number, validators: number, totalTransactionCount: number}}
+ * @param {boolean} visible                   Only fire new queries when visible.
+ * @param {number} performanceUpdateSeconds   Delay before next query.
+ * @param {number} sampleHistoryHours         How many hours (60min.) the query should go back.
```
tigarcia, line 75: This is a bit confusing.  Maybe a better variable name here to make it clear that the sampleHistoryHours is only for the super majority check and not for the tps check.

src/hooks/useTransactionStats.js
```diff
@@ -1,68 +1,81 @@
...
     ? new AbortController()
     : { abort: noOp, signal: {} };
 
-const isDevelopment = process?.env?.NODE_ENV === "development" || false;
+/**
+ * Fetches the superminority count.
+ *
+ * @returns {Promise<number>} The count of superminority validators.
+ */
+export const fetchSuperminority = async () => {
```
tigarcia, line 30: The logic here looks good to me. One potential issue with the sort.

src/hooks/useTransactionStats.js
```diff
@@ -143,35 +145,37 @@ export const useTransactionStats = ({
...
     return () => {
       abortController.abort();
-      // Clear interval to prevent multiple calls to setInterval.
-      if (interval) {
-        clearInterval(interval);
-      }
+      clearInterval(interval);
     };
   }, [
     visible,
```
tigarcia, line 184: Does the superminority ever get changed based on time? I guess the visibility would likely change at some point.

tigarcia (commented):
Looks good overall. I added a few comments to clarify some of the changes.

src/hooks/useTransactionStats.js
catalinred: I see, in this PR I haven't touch `sampleHistoryHours` at all so maybe we can address this onwards in a different PR.

src/hooks/useTransactionStats.js
catalinred: Thanks for the suggestion, I updated the code. Also I updated the comment to reflect the ascending order.

src/hooks/useTransactionStats.js
catalinred: I don't think we want the superminority to update it periodically like it does for other stats.

The above `setInterval` function purpose is to help update  this block below.
image

I remember in one of my previous iterations, I had the current superminority calculated and updated periodically like the other stats, but this quickly exhausted my free tier RPC plan.

tigarcia (approved):
Looks good 👍

PR #37, opened 2024-09-10
feat: support strings in tabs `items`

nickfrosty:
added support for passing a string value for the `items` inside the `Tabs` component

this will allow passing a comma-separated list of the tab labels within dev content files

note: this is to support fixing the render errors with several open issues in the dev-content repo: https://github.com/solana-foundation/developer-content/issues/455
these issues are being caused by crowdin altering the content files in an odd way, specifically when an MDX component has an attribute with an array value, crowdin prepends a `\` in front of the components opening tag. breaking the parser....


catalinred (approved):
Thank you Nick!

PR #38, opened 2024-09-11
fix: deduplicate filters

nickfrosty:
filters can appear multiple times when they are listed in the dev content api with different cases

image

- added a case insensitive deduplication to each of the filters


catalinred (approved):
Lgtm!

Issue #58, opened 2024-10-07
Bug: YouTube subscriber count displaying as zero on the Solana Community page


devansh-m12:
## Steps To Reproduce
1. Go to the Solana website's [Community page](https://solana.com/community).
2. Scroll to the Wayfinding section displaying the YouTube count.
3. Observe the YouTube count displayed as zero.

## The current behavior
The YouTube count is showing as zero on the website, despite the actual number of subscribers on the YouTube channel being higher.
<img width="1398" alt="Screenshot 2024-10-07 at 1 12 26 PM" src="https://github.com/user-attachments/assets/9466ebac-5a84-4f59-84a0-6ab05b1bb1d2">

## The expected behavior
The YouTube count should display the correct number of subscribers from the official Solana YouTube channel.

i have also fixed this, please allow so i can submit a pr

PR #59, opened 2024-10-07
fix: cheerio import and yt scrape data

devansh-m12:
This fixes https://github.com/solana-foundation/solana-com/issues/58

here the proof : 
Screenshot 2024-10-07 at 1 18 00 PM

### DIFF
--- a/src/utils/followerFunctions.js
+++ b/src/utils/followerFunctions.js
@@ -68,6 +68,29 @@ const scrapeYoutubeSubscriberCount = async () => {
   return 0;
 };
 
+const getYoutubeSubscriberCount = async () => {
+  const channelId = process.env.YOUTUBE_CHANNEL_ID
+  const YOUTUBE_API_KEY = process.env.YOUTUBE_API_KEY
+  try {
+    const response = await fetch(`https://www.googleapis.com/youtube/v3/channels?part=statistics&id=${channelId}&key=${YOUTUBE_API_KEY}`);
+
+    if (!response.ok) {
+      throw new Error(`Error fetching data: ${response.statusText}`);
+    }
+
+    const data = await response.json();
+
+    if (data.items && data.items.length > 0) {
+      return parseInt(data.items[0].statistics.subscriberCount, 10)/1000;
+    }
+
+    throw new Error('Channel statistics not found');
+  } catch (error) {
+    console.error('Error getting subscriber count:', error.message);
+    throw error;
+  }
+};
+
 const scrapeVKontakteFollowerCount = async () => {
   const followerCountTag = await scrapeUrlForTag(
     `https://vk.com/solanarus`,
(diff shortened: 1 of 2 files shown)


devansh-m12:
We're already using Google's APIs to fetch YouTube playlists and videos, so why not use the same APIs to get the subscriber count instead of scraping the data, which avoids the uncertainties of updates?

src/utils/followerFunctions.js
```diff
@@ -1,4 +1,4 @@
-import cheerio from "cheerio";
+import * as cheerio from "cheerio";
```
catalinred, line 1: Are these change needed if `scrapeYoutubeSubscriberCount` is not used anymore?

If `scrapeYoutubeSubscriberCount` is getting superseded by `getYoutubeSubscriberCount` let's remove the former function completely

src/utils/followerFunctions.js
```diff
@@ -54,17 +54,45 @@ const scrapeYoutubeSubscriberCount = async () => {
...
   }
   return 0;
 };
 
+const getYoutubeSubscriberCount = async () => {
+  const channelId = process.env.YOUTUBE_CHANNEL_ID;
+  const YOUTUBE_API_KEY = process.env.YOUTUBE_API_KEY;
+  try {
+    const response = await fetch(
+      `https://www.googleapis.com/youtube/v3/channels?part=statistics&id=${channelId}&key=${YOUTUBE_API_KEY}`,
```
catalinred: Can you move `baseURL` outside as global const to use in both `getYTVideos` and newly added `getYoutubeSubscriberCount `?

catalinred (changes requested):
Thanks for working on this @devansh-m12
This looks good overall, I added a couple of comments below

src/utils/followerFunctions.js
devansh-m12: These changes are essential for `scrapeUrlForTag`, which is used to extract both the `Meetup member count` and `VK follower count`. As seen in the image, the fixes for the `Meetup member count` are now functioning correctly. However, there is still an issue with the `VK follower count`, which I plan to address in a separate PR. Additionally, as per your instructions, I will remove the `scrapeYoutubeSubscriberCount` function.

catalinred (approved):
Lgtm, thanks for working on it!

devansh-m12:
@catalinred Thanks! On a separate note, the VK followers count isn't working, and it seems like much of the data in the community stats section is hardcoded. I'd be interested in working on that.

catalinred:
@devansh-m12 I will have this branch merged next and fwiw I just added the #hacktoberfest topic label to this repo ;) I think this initiative is great, thanks for the suggestion.

Some of the community stats are hardcoded and could definitely use some attention. Feel free to open an issue or submit a PR, and we can continue the discussion there on how to address it.

PR #66, opened 2024-10-21
Add Solana Explorer to Header Menu

metasal1:
added Solana Explorer to the  Header Menu under Network / Inspect and provided 18 language translations


catalinred:
Good call @metasal1, thanks for working on this.

Can you please update the description and translation to just `Blocks, accounts, transactions, programs and tokens.`? Also it would be great if you could add the missing translations, at first sight I think it's `ko` and `hi`.

catalinred (changes requested):
Please open a separate PR to suggest the changes for the Fuse wallet if needed.

catalinred:
> Please open a separate PR to suggest the changes for the Fuse wallet if needed. Also opening an issue beforehand would be recommended.

@metasal1 so your newly opened PRs are duplicated and they contain both the Solana Explorer addition to the menu and the Fuse Wallet updates. Can you review their changes and decide how to isolate the changes for each issue and PR?

PR #93, opened 2024-11-06
Fix courses link

mikemaccana:
Currently goes to `/developers#courses`, but courses are on `/developers/courses`


catalinred (approved):
Thank you @mikemaccana

Issue #61, opened 2024-10-08
Improve 'Solana Changelog' Section: Display Latest Episode Thumbnail and Format Recent Videos List


devansh-m12:
## Steps To Reproduce
1. Go to the 'Solana Changelog' section (as seen in the attached screenshot).
2. Notice that instead of displaying the latest episode thumbnail, the section shows a text description that appears messy and unformatted.

## The Current Behavior
- The 'Solana Changelog' section displays a simple text description with broken links and formatting issues.
- The 'Latest Episode' button does not highlight the latest episode effectively, missing an engaging visual (e.g., a thumbnail).

## The Expected Behavior
- The 'Solana Changelog' section should prominently display a thumbnail for the latest episode, making it visually engaging and informative.
- Below the thumbnail, there should be a well-formatted list of recent episodes/videos, providing an easy-to-navigate experience.
- The button should focus on directing users to the latest episode with a clear call-to-action.
- The description text should be properly formatted with working links and clean layout.
  
## Proposal
I propose redesigning the 'Solana Changelog' section to focus on the following:
- Add the latest episode thumbnail in focus.
- Provide a formatted list of recent videos below the thumbnail.
- Ensure clean and working links in the description.
  
This behavior can be observed in the `/developers` section as a reference, where the layout is more user-friendly and organized. I would be happy to work on this enhancement and discuss further.

**Screenshot for reference:**
<img width="1453" alt="Screenshot 2024-10-08 at 10 26 52 AM" src="https://github.com/user-attachments/assets/3cb2c6c0-6233-457b-9ab0-c64a6ac05973">

catalinred:
Thanks for the suggestion @devansh-m12 

At the moment, updating the UI/UX for this section is not a priority. We intentionally chose not to display the latest episode's thumbnail primarily for design reasons, as the episode thumbnail may not always be the most visually appealing option. Additionally, links are not rendered within the description to prevent potential code injection.

nickfrosty:
@catalinred since I am the one crafting the changelog video descriptions, I will add a short paragraph at the top instead of just listing things like we currently do. This way the details displayed on the site are more coherent. I will take some time next week to update the last few video descriptions as well, and make sure to do this going forward :)

PR #117, opened 2024-11-22
fix: changelog description

nickfrosty:
updated the changelog description parser to handle no description and splitting based on the video description containing `---` as a separator

for future changelog video descriptions (and several past ones), we will start adding a summary paragraph at the top followed by a `---` separator so solana-com can better display

note: for other descriptions without this separator, similar video description behavior as before will occur, but just a little shorter

Fixes: #61

### DIFF
--- a/src/components/developers/sections/DevelopersDocumentsSection/DevelopersChangelog.jsx
+++ b/src/components/developers/sections/DevelopersDocumentsSection/DevelopersChangelog.jsx
@@ -14,7 +14,14 @@ export default function DevelopersChangelog({ latestVideo }) {
     <div className={styles["changelog"]}>
       <h3 className={styles["changelog__title"]}>Solana Changelog</h3>
       <p className={styles["changelog__description"]}>
-        {truncateTextByWord(latestVideo.snippet?.description, 240, "...")}
+        {truncateTextByWord(
+          (
+            latestVideo.snippet?.description ||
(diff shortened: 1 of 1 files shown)


src/components/developers/sections/DevelopersDocumentsSection/DevelopersChangelog.jsx
```diff
@@ -14,7 +14,14 @@ export default function DevelopersChangelog({ latestVideo }) {
     <div className={styles["changelog"]}>
       <h3 className={styles["changelog__title"]}>Solana Changelog</h3>
       <p className={styles["changelog__description"]}>
-        {truncateTextByWord(latestVideo.snippet?.description, 240, "...")}
+        {truncateTextByWord(
+          (
+            latestVideo.snippet?.description ||
+            "Latest changes for the Solana blockchain"
```
catalinred, line 20: lgtm, my only feedback is that this string is not localized - just in case a video has no description.
Not sure if we want to add this to the locale files, but maybe we don't need the OR conditional here at all?

nickfrosty: This string should only even be displayed if the YouTube API call fails for some reason.
Plus we have no way to localize the content in the video description itself

catalinred (approved):

PR #121, opened 2024-11-23
[courses] added hero to courses page

nickfrosty:
## Summary of changes

- refactored the developer hero section to accept labels and button info via props
- added a hero section to the courses with some CTAs

note: instead of using the same "geometry" images at the /developers page, I used the validators geometry to add more noticeable visual difference between the two pages (/developers and /courses)

edit: the diff on the courses pages looks far more dramatic of a change than it actually is since I had to change the Layout to use the generic one to avoid clipping the courses hero. so to follow the /developers page, I wrapped the content in the same `div`


catalinred (approved):
Lgtm!

PR #149, opened 2025-01-19
feat: Add clipboard fallback for CopyToClipBoardButton

VolodymyrBg:
### Problem

TODO

### Summary of Changes

Implemented a fallback copy mechanism using document.execCommand('copy') for browsers that don't support the modern Clipboard API. This ensures the copy functionality works across a wider range of browsers.

- Added fallback implementation using document.execCommand
- Improved error handling and cleanup
- Removed TODO comment

### DIFF
--- a/src/components/shared/MarkdownRenderer/components/CopyToClipBoardButton.tsx
+++ b/src/components/shared/MarkdownRenderer/components/CopyToClipBoardButton.tsx
@@ -20,16 +20,32 @@ export function CopyToClipBoardButton() {
 
   const copyToClipboard = useCallback(async () => {
     try {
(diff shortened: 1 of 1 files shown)


catalinred:
Thanks for looking into this but as far as I know there is an impressive [browser support](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard#browser_compatibility) for the navigator clipboard property already.

This [comment](https://github.com/solana-foundation/solana-com/pull/149/files#diff-4c2c27d4ca067d710c44ca1056af2bdc5ae51ada6939af2ee6d68680f7934510L24) is misleading and was added long time ago.

PR #111, opened 2024-11-15
Add Fumadocs

pomber:
This is a draft PR showing how to use fumadocs as the docs framework. 
- currently only applied to `/docs` path (with placeholder content)
- the rest of the website remains using the pages router without any changes

TODO
- there are some issues with the postcss plugins I still need to fix (I disabled them for now)
- locale routing isn't included so far, the app router only works for English
- added some `@ts-ignore` annotations that can be removed after the old docs framework is replaced (because of some conflicts between different versions of `mdx/types`)

RE Tailwind duplication: 
fumadocs components ship with unprefixed classes, while solana-lib use the `tw` prefix, this means that we need to include two tailwind stylesheets (only for the routes that use fumadocs),  we can still use something like `cssnano` to remove any duplicated CSS (like tailwind's preflight styles)

PR #154, opened 2025-01-24
Fix SEO and i18n for some /developers pages

catalinred:
### Problem
Related #111 

### Summary of Changes
- Localize some missing titles and descriptions
- Update the OGP title and description fields for `/developers/courses`, `/developers/guides` and `/developers/resources`. Previously the OGP fields were set to global site title and description.

image

### DIFF
--- a/src/app/[locale]/developers/courses/[course]/[lesson]/page.tsx
+++ b/src/app/[locale]/developers/courses/[course]/[lesson]/page.tsx
@@ -4,6 +4,7 @@ import { mdxComponents } from "@/app/mdx-components";
 import { BlogPage } from "@/app/components/blog-page";
 import { getAlternates } from "@/i18n/routing";
 import { getUrlWithoutLocale } from "@/app/sources/utils";
+import { ResolvingMetadata } from "next";
 
 type Props = {
   params: Promise<{ course: string; lesson: string; locale: string }>;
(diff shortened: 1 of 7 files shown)


catalinred:
@pomber for vis.

pomber:
To avoid losing the other openGraph tags (set by `[locale]/layout.tsx`) you can spread the parents metadata. 

Here an example from `src\app\[locale]\developers\page.tsx`:

```tsx
 function generateMetadata(
  { params }: Props,
  parent: ResolvingMetadata,
) {
  const { locale } = await params;
  const { t } = await serverTranslation(locale);
  const { openGraph } = await parent;
  return {
    title: t("developers.title"),
    description: t("developers.description"),
    openGraph: {
      ...openGraph,
      title: t("developers.title"),
      description: t("developers.description"),
    },
    alternates: getAlternates("/developers", locale),
  };
}
```

pomber:
> In all cases we want the OGP's title and description to be the same as title and description so perhaps we need to update layout.tsx to accommodate that instead.

My understanding was that all opengraph clients default to the page's client and description when the opengraph tags aren't set.

Anyway, I can create a `getMetadata` helper function to avoid all the repeated logic. I'll do it in a new PR. Let me know if you need it for today or it can wait.

pomber:
Moved to #157

catalinred:
Superseded by https://github.com/solana-foundation/solana-com/pull/157

PR #158, opened 2025-01-27
Add `/content` folder to Crowdin

pomber:
### Problem

Recently migrated content files aren't managed by Crowdin.

### Summary of Changes

Added new sources to `crowdin.yml`

---

This is what I think we need based on the [docs](https://support.crowdin.com/developer/configuration-file/). Since there's no way to test this before committing, maybe we should try using a subfolder first instead of the entire `/content` folder, something like `/content/docs/intro`.

One concern I have is how Crowdin handles translated files once they're committed. It’s unclear if it will ignore files like `/content/index.es.mdx` or mistakenly treat them as sources.

### DIFF
--- a/content/docs/intro/wallets.es.mdx
+++ b/content/docs/intro/wallets.es.mdx
@@ -1,0 +1,60 @@
+---
+title: Wallets
+h1: Guía de Wallets de Solana
+---
+
+Este documento describe las diferentes opciones de wallet disponibles para los usuarios de Solana que quieren ser capaces de enviar, recibir y interactuar con SOL tokens en la blockchain de Solana.
+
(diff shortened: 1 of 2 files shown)


catalinred (commented):
@pomber I like the idea of trying the synchronisation with a subfolder first. This way, having already access to Crowdin, you can test the actual sync and iterate afterwards.

I also believe you will need at least a `/content/index.[lang].mdx` file to test with and see whether Crowdin recongnize the two letters code language when setting up the new folder.

catalinred (approved):
@pomber lgtm, will merge this so you can continue testing this out in Crowdin.

PR #159, opened 2025-01-27
Lowercase URLs

pomber:
### Problem

- URLs should be all lowercase (to match the URLs before migration)

### Summary of Changes

- rename content filenames to be lowercase (only rpc folder had uppercase filenames)
- add middleware to redirect to the lowercase version of any URL


catalinred (approved):
This looks good to me, ty @pomber!

PR #184, opened 2025-02-01
replace try_to_vec with try_to_vec_with_schema

Levi0804:
### Problem
In `borsh` v1.5.5 the `try_to_vec` method is no longer provided by `BorshSerialize`. Instead we can use `try_to_vec_with_schema` by deriving `BorshSchema` on the struct.

### Summary of Changes
- Replaced `try_to_vec` with `try_to_vec_with_schema`
- Added`#[derive(BorshSchema)]` on `NewAccount` struct

### DIFF
--- a/content/guides/getstarted/intro-to-native-rust.mdx
+++ b/content/guides/getstarted/intro-to-native-rust.mdx
@@ -26,7 +26,7 @@ program. Here is the program on
 [Solana Playground](https://beta.solpg.io/661058a6cffcf4b13384d02a).
 
 ```rust title="lib.rs"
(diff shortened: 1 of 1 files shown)


ZYJLiu:
It seems like the [schema module](https://docs.rs/borsh/latest/borsh/schema/index.html) from borsh crate behind an [unstable_schema](https://github.com/near/borsh-rs/blob/master/CHANGELOG.md#stability-guarantee) feature flag.

Also, the program changes results in the test file no longer working when the instruction is invoked on Solana Playground.
What would be the JS/TS equivalent when using BorshSchema?
The Solana Playground shared link would also need to be updated.

PR #190, opened 2025-02-04
Unify opengraph endpoint

pomber:
### Problem

- Build errors related to the opengraph endpoint

### Summary of Changes

- Replaced `fumadocs-mdx` with a fork: `fs-mdx`
- Removed the `content/sync` task (now part of `fs-mdx`)
- Unified and optimized the opengraph endpoint

---

This should also improve the dev performance for `courses` and `guides`, but the major improvement will come with the release of [async sources](https://x.com/fuma_nama/status/1885696201640775789).


ZYJLiu:
could you provide more context regarding the benefit of the `fs-mdx` fork and the expected dev performance improvement?

pomber:
> could you provide more context regarding the benefit of the `fs-mdx` fork and the expected dev performance improvement?

So, the way that `fumadocs-mdx` works is by creating a `./source/index.js` file that imports all the mdx files, importing that file is very slow for the amount of files we have. 

In #111, I added a "hack" (the script in `/content/sync.mjs`) that splits that file into one file for each collection (docs, cookbook, courses, guides). 

`fs-mdx`, it's the same as `fumadocs-mdx` but with two improvements:
- the collection-splitting is integrated inside the library instead of being a hack
- it creates extra files with just the frontmatter of each collection, for example `./source/cookbook.fm.js`, which are much cheaper to import. This PR uses those for the `opengraph` endpoint.

ZYJLiu:
does this address the vercel [sigkill error](https://vercel.com/guides/troubleshooting-sigkill-out-of-memory-errors) that we're occasionally seeing on deployments?

if I'm understanding correctly, the performance is appoximately the same (it wasn't noticable faster when this branch running locally) with additional optimization for the opengraph endpoint (for images?) 

is `fs-mdx` intented a temporary solution, which will be swapped out when fumadocs releases async sources, or a permanent thing?

catalinred (approved):
I do like how the OGP codebase for images generation is better organized and merged within `src/app/opengraph/developers/[...slug]/route.ts`

PR #204, opened 2025-02-08
docs: revamped README with improved layout and badges

NikolaiKryshnev:


catalinred (commented):
While I do like some bits such as highlighting of stars and forks, we won't be merging this for now for a couple of reasons.

Also in the future please add a description to your PR and explain what issue you are trying to address.

PR #214, opened 2025-02-12
`new_readonly` is not not writable account

vikiival:
### Problem

Went through https://solana.com/developers/guides/getstarted/how-to-cpi

and found out that docs states:

```
AccountMeta::new_readonly(account4_pubkey, true),  // writable, signer
```

and by definition of `readonly` account cannot be writable

### Summary of Changes

renamed comment from `// writable, signer` to `// not writable, signer`

Fixes #


ZYJLiu (approved):

ZYJLiu:
thank you!

PR #216, opened 2025-02-12
fix spelling errors

Vitaliyr888:
Lets - Let's 

just like - just is like

### DIFF
--- a/content/cookbook/tokens/manage-wrapped-sol.mdx
+++ b/content/cookbook/tokens/manage-wrapped-sol.mdx
@@ -5,7 +5,7 @@ description:
   adding balance through SOL transfers or token transfers."
 ---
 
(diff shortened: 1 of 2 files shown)


content/cookbook/tokens/manage-wrapped-sol.mdx
```diff
@@ -5,7 +5,7 @@ description:
   adding balance through SOL transfers or token transfers."
 ---
 
-Wrapped SOL just like any other token mint. The difference is using `syncNative`
+Wrapped SOL just is like any other token mint. The difference is using `syncNative`
```
ZYJLiu:
```suggestion
Wrapped SOL is just like any other token mint. The difference is using `syncNative`
```

ZYJLiu (changes requested):

ZYJLiu:
thank you!

PR #226, opened 2025-02-15
fix: broken web3.js V2 react example link

jim4067:
### Problem
Broken link for react app example using web3.js V2


ZYJLiu:
thank you!

Afeter8 (changes requested):
Glg

PR #240, opened 2025-02-21
Docskit components and annotations

pomber:
### Summary of Changes

- Add [docskit](https://docskit.codehike.org/) components and annotations (except Language Switcher and Terminal, I'll add them later)
- Add a hidden docs page with usage examples at /docs/docskit


package.json
```diff
@@ -13,13 +13,17 @@
...
+    "@radix-ui/react-tabs": "^1.1.3",
+    "@radix-ui/react-tooltip": "^1.1.8",
     "@sindresorhus/slugify": "^2.2.1",
     "@solana-foundation/solana-lib": "^2.40.2",
     "@typeform/embed": "^5.1.0",
     "bootstrap": "^5.3.3",
     "cheerio": "^1.0.0",
+    "class-variance-authority": "^0.7.1",
     "classnames": "^2.5.1",
+    "clsx": "^2.1.1",
```
catalinred: @pomber Doesn't the existing `classnames` package handle conditional `className`'s in the same way?
Maybe [src/app/components/utils.tsx](https://github.com/solana-foundation/solana-com/pull/240/files#diff-bfe9b93aeb9a11bf677484fba342706c880c8a4b7167a2a3c0871097ee335f37) can be refactored to use `classnames` instead in this case and have `clsx` removed.

pomber: yes, I'll change it

ZYJLiu (approved):

PR #243, opened 2025-02-24
Set HTML lang attribute

pomber:
### Problem

html tag is missing the lang attribute (in pages router)

### Summary of Changes

Use the locale prop in _document.js to set the lang attribute


catalinred (approved):
Thank you @pomber

PR #256, opened 2025-02-25
Remove invalid search params

pomber:
### Problem

During client-side navigation nextjs thinks that href="solutions/token-extensions" should go to the catch-all "[locale]/[...slug].js" so it sends the URL params "locale=solutions&slug=token-extensions"

### Summary of Changes

Remove the locale and slug params in the middleware. We only do it when the locale param isn't a valid locale, just to be extra safe.


catalinred (approved):
Thanks, @pomber! 
This looks good to me. I'll run a few more tests and merge it afterward.

PR #283, opened 2025-03-03
docs: update load-default-keypair cookbook snippets

jim4067:
### Summary of Changes
- add rust clients examples
- add legacy web3.js example


ZYJLiu:
thank you! btw `web3.js` v2 is now published as `@solana/kit`

PR #294, opened 2025-03-06
New RPC docs layout

pomber:
This is how it looks so far:

https://github.com/user-attachments/assets/d5216f30-5431-4133-bae9-8e2026ee0b8a


ZYJLiu:
Thanks Rodrigo! The updated UI looks great!
It looks like all the http methods are migrated.
Is this ready to merge or are you still making updates?

PR #311, opened 2025-03-12
docs: add Inkeep Searchbar

ZYJLiu:
- add Inkeep Searchbar, limited to only /docs and /developer pages

### DIFF
--- a/src/app/components/inkeep/inkeep-searchbar.tsx
+++ b/src/app/components/inkeep/inkeep-searchbar.tsx
@@ -1,0 +1,15 @@
+"use client";
+import dynamic from "next/dynamic";
+import { useInkeepConfig } from "./useInkeepConfig";
+
+const SearchBar = dynamic(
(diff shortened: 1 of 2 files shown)


catalinred (commented):
We need to prevent the current FOUC effect introduced by the new Inkeep component.

I made a video that shows this flashing issue, and you can see how the header and therefore the page jumps because the search `input` is not visible and styled by default and it gets rendered only afterward.

Maybe this don't need to use dynamic import.

https://github.com/user-attachments/assets/e2aae372-25b7-4e3a-bbe8-41810c31e0d4

ZYJLiu:
@catalinred updated to use same styling as existing searchbar and fixed issue you mentioned


https://github.com/user-attachments/assets/a095b5a7-786f-46b3-9083-c73628b874c8

catalinred (approved):
Thank you @ZYJLiu, this looks good to me!

PR #312, opened 2025-03-13
Correct `commitment` property of `getBlockCommitment`

steveluscher:
### Problem

The `commitment` array refers to how much stake has lent 1-32 confirmations to a block. The array is 32 items long. This would imply that the depth ranges from 0 to 31 (zero-indexed) – not 0 to 32 as implied by the docs.

### Summary of Changes

Remove the “+ 1”


ZYJLiu:
thank you!

ZYJLiu (approved):

PR #316, opened 2025-03-14
Remove background from header buttons

vladan345:
### Problem
Small design changes

### Summary of Changes
Request was to remove background color from 2 buttons located in the Desktop header

Fixes #
Changed style in Sass file for those buttons and added colored bottom border on hover


catalinred (approved):
Lgtm now, ty @vladan345

PR #341, opened 2025-03-20
Add CTA to Accelerate Header

vladan345:
### Problem
CTA was requested to be in Header

### Summary of Changes
Added the button and added styles

Fixes #


src/components/accelerate/AccelerateHeader.jsx
```diff
@@ -69,6 +69,14 @@ const Header = () => {
                 </AccelerateButton>
               </Link>
             </li>
+            <li className="nav-item">
+              <Link
+                href="https://lu.ma/accelerate2025"
+                className={styles["cta"]}
+              >
+                Get tickets
```
catalinred: This string needs to be moved to the locale file https://github.com/solana-foundation/solana-com/blob/main/public/locales/en/common.json#L1027-L1033

src/components/accelerate/AccelerateHeader.module.scss
```diff
@@ -5,3 +5,44 @@
...
+  display: inline-block;
+  @media screen and (max-width: 992px) {
+    margin-left: 0;
+    margin-top: 16px;
+  }
+}
+.cta::after {
+  color: black;
+  background: white;
+  content: "Get tickets";
```
catalinred: Why is the "Get tickets" string repeated within a pseudo-element? 

- This it will break in other languages -  when the above string is moved to common.json.
- Looking at the styling of the button I don't think this pseudo-element is needed at all - at first sight the same effect can be achieved without using pseudo-elements.

Additionally, the button’s overall style feels oversized and somewhat off - was this intentional?

catalinred (changes requested):

src/components/accelerate/AccelerateHeader.module.scss
vladan345: I wouldn't say that the button is oversized, it matches the typography of the links next to it, plus padding to have a background. Plus, it's supposed to be a CTA so it should attract user's attention.

Regarding pseudo-element I agree. The same effect can be done in jsx, the reason i need two of the same button is transition. I'll update it with locale as well.

catalinred (approved):

PR #345, opened 2025-03-21
Feat/accelerate sponsors

vladan345:
### Problem
Request to make Sponsors component for Accelerate page

### Summary of Changes
Added new custom component that accepts array of logos and names

Fixes #


catalinred (changes requested):
@vladan345 please create a new Sponsors PR, based on main, and `cherry-pick` the specific Sponsors changes only

PR #364, opened 2025-03-25
Make `filters` required in `getTokenAccountsByDelegate()`

steveluscher:
Reference: https://github.com/anza-xyz/agave/blob/091a7b4084f9397b803a4f19dd5c2a5446c957fb/rpc/src/rpc.rs#L3461


ZYJLiu:
updated to param to "required" on new rpc page
https://solana.com/docs/rpc/http/gettokenaccountsbydelegate

PR #365, opened 2025-03-25
Note that reward data can't be refused when using `getTransaction()`

steveluscher:
I'm pretty sure that the `getTransaction()` API does not have a `rewards` property like `getBlock()` does.


ZYJLiu:
transaction metadata description updated with correction and moved to:
https://solana.com/docs/rpc/json-structures#transaction-status-metadata


getBlock and getTransaction pages now link to the metadata description
https://solana.com/docs/rpc/http/getblock
https://solana.com/docs/rpc/http/gettransaction

PR #367, opened 2025-03-26
Mirror TS Code Exec

hwsimmons17:
### Problem
Developers want to be able to run code examples in the browser to better understand what is going on with the code.

### Summary of Changes
Added `flags="r"` flag to the `<CodeTabs>` component allowing any CodeTabs component to become a runnable componeent

Fixes #
Went through each of the code examples ensuring they work.

TODO:
* Add Rust Support
* Add pre-existing programs so functionality that requires existing programs such as on /developers/cookbook/programs/clock actually work
* Ensure Gill is working properly. I had to add ts-ignore in places and such... honestly getting it to work was quite a slog and I'm not sure Solana should be recommending it as the default client yet, but there is more work that needs to be done there to ensure that works seamlessly.

### DIFF
--- a/src/app/components/code/code.tsx
+++ b/src/app/components/code/code.tsx
@@ -15,6 +15,8 @@ import { wordWrap } from "./word-wrap";
 import { MultiCode } from "./code.client";
 import { tooltip } from "./tooltip";
 import { tokenTransitions } from "./token-transitions";
+import { getMirrorInstance } from "@/utils/mirror";
+import { RunableCode } from "./code.runable";
 
 export async function Code(props: {
   codeblocks: RawCode[];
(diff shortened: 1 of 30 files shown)


ZYJLiu:
thank you!

ZYJLiu:
There seems to be an issue with the setup. Pages set up with mirror are throwing this error on the vercel preview.
It might have something to do with server side rendering

CleanShot 2025-03-25 at 21 53 26@2x

ZYJLiu:
Here's the vercel log
CleanShot 2025-03-25 at 21 58 00@2x

ZYJLiu:
tried it out locally and the mirror service works!

However, when I first open the browser on localhost, I get this error:
CleanShot 2025-03-26 at 15 36 19@2x

I'm able to just close an then navigate to http://localhost:3000/docs, where then the error doesn't get thrown.
Are you getting same error and any idea how to fix?

hwsimmons17:
It's something I need to fix tonightSent from my iPhoneOn Mar 26, 2025, at 2:39 PM, John ***@***.***> wrote:﻿
  tried it out locally and the mirror service works!
However, when I first open the browser on localhost, I get this error:
***@***.*** (view on web)
I'm able to just close and then navigate to http://localhost:3000/docs, where then the error doesn't get thrown.
Are you getting same error and any idea how to fix?—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



ZYJLiu left a comment (solana-foundation/solana-com#367)
tried it out locally and the mirror service works!
However, when I first open the browser on localhost, I get this error:
***@***.*** (view on web)
I'm able to just close and then navigate to http://localhost:3000/docs, where then the error doesn't get thrown.
Are you getting same error and any idea how to fix?

hwsimmons17:
It's something I need to fix tonight. You can just exit out of the warning and it will work if you want to text more

src/app/[locale]/layout.tsx
```diff
@@ -25,6 +26,7 @@ export default async function RootLayout({ children, params }: Props) {
   const { resources } = await initTranslations(locale, namespaces);
   const googleTagManagerID = config.siteMetadata.googleTagManagerID;
   const builderLocale = locale == "en" ? "Default" : locale;
+  setUserID();
```
Comment (pomber) [inline src/app/[locale]/layout.tsx]: I think using cookies here will opt-out the page from the static build, is it possible to delay this until the user clicks the "run" button?

src/app/components/code/code.tsx
```diff
@@ -77,6 +91,27 @@ export async function toCodeGroup(props: {
       const { flags, title } = extractFlags(tab);
       const tabOptions = flagsToOptions(flags);
       const options = { ...groupOptions, ...tabOptions };
+      // get the user_id from the users cookies
+
+      if (groupOptions.runable) {
+        const { mirrorUrl, wsMirrorUrl } = await getMirrorInstance();
```
pomber: same with this, is it possible to move it to the "run" handler?

hwsimmons17: That is a great point! I will change that.

hwsimmons17: We can do that!

hwsimmons17: The only thing to note is that you would lose the url being loaded on page load, it would show localhost:8899 until the user clicked run. Not sure if that is an issue from a product perspective

hwsimmons17: I personally care about that from an attribution perspective

ZYJLiu:
Vercel Preview is not working
CleanShot 2025-03-28 at 19 15 55@2x

Vercel logs:
CleanShot 2025-03-28 at 19 16 46@2x

src/app/components/code/code.tsx
ZYJLiu: for attribution, is it not sufficient to have just have the existing "powered by mirror" link next to the copy button on the code snippet, and also noted on the output accordion?

hwsimmons17:
@ZYJLiu Are you running the latest version of the code? I am running it here: https://solana-com-1ucd.vercel.app/developers/cookbook/transactions/send-sol and it's working. handle when code type is ts not typescript is the latest commit

ZYJLiu:
thats the latest preview deployment from the last commit

src/app/components/code/code.tsx
hwsimmons17: If you copy the code, it would be nice for it to have the mirror link directly in it! That is the attribution I'm talking about...

ZYJLiu:
I'm seeing the same error on the link you're link as well.
https://solana-com-1ucd.vercel.app/developers/cookbook/transactions/send-sol:

CleanShot 2025-03-28 at 19 22 59@2x

hwsimmons17:
I know why, it's because my site already has cookies, so it doesnt fail... I will make that change that @pomber recommended. Give me 30.

hwsimmons17:
It's working!

ZYJLiu (approved):

PR #375, opened 2025-03-27
copy changes + button fix

ybhrdwj:
1. Fixed a bunch of copy + links
2. Fixed a broken get started button that was breaking due a conflicting css property


catalinred (approved):
ty @ybhrdwj

PR #378, opened 2025-03-27
Replace Algolia searchbar with Inkeep searchbar

ZYJLiu:
- unify searchbar across solana-com using Inkeep search for all pages


ZYJLiu:
@catalinred I've updated solana-com where the Inkeep searchbar is used for all pages.
The preview deployment seems to be working. Are there any other things that need to be changed or removed?

The `@docsearch/react` is still included since the [Inkeep searchbar component](https://github.com/solana-foundation/solana-com/blob/main/src/app/components/inkeep/inkeep-searchbar.tsx) was using the algolia searchbar css.
I think to remove the extra dependency would just be updating the searchbar styling. 

The current implmentation was added to avoid the FOUC while the default Inkeep searchbar component loaded.
Do you want to keep it as is or update the styling to remove the `@docsearch/react` dependency?

catalinred:
thank you @ZYJLiu!
trying to help so I also pushed a couple of changes to remove obsolete references and more.

I think it's ready to go live, let me know if you notice anything off - I removed old scss files imports, restyled the searchbox and remove all the docbook* references.

PR #407, opened 2025-04-10
docs: update and move token page

ZYJLiu:
- move `/docs/core/token` to `/docs/token`
- fix links and simplify content


neerrrajj:
@ZYJLiu, do you plan to add the spl-token cli examples that were there in 'core/token' to the new token page? they were useful

ZYJLiu:
sure I'll add them back

PR #409, opened 2025-04-10
docs: add rust snippets for token examples

jim4067:
### Summary of Changes

docs: add Rust snippets for fungible token examples


ZYJLiu:
thank you!

PR #415, opened 2025-04-15
Make `/content/docs` support locale subfolders

pomber:
### Summary of Changes

Support locale subfolders in `/content/docs`.

For example, if there's a `/content/docs/es/intro/installation.mdx` it will be used for the `es` locale, otherwise we fallback to `/content/docs/en/intro/installation.mdx`.


ZYJLiu:
This is great, thank you! 
It looks like the images that would be on https://solana.com/developers/courses are not working on the vercel preview
CleanShot 2025-04-15 at 04 32 18@2x

pomber:
> It looks like the images that would be on https://solana.com/developers/courses are not working on the vercel preview 

I think that happens with all the PR previews, right?

ZYJLiu:
Just checked a preview for a different PR and it looks like you're right

ZYJLiu (approved):

PR #442, opened 2025-04-28
docs: Solana kit.js, web3.js and rust snippets

jim4067:
### Summary of Changes
Add Solana kit.js, web3.js and rust snippets to RPC docs


ZYJLiu:
This is great thank you! 
Out of the ones that have been updated, some don't have all snippets.
Was there a reason some were skipped or just missed them?

jim4067:
Some, especially in the old web3.js `connection` object don't have those methods, so I ended up skipping them.

Issue #288, opened 2025-03-05
Misc Codehike Issues


ZYJLiu:
## WithMentions

`!mention` only works when labels are `one` or `two`, doesn't work with arbitrary label

````mdx

<WithMentions>
- Works with [one](mention:one)
- Works with [two](mention:two)
- Does not work with [three](mention:three)
- Does not work with arbirary label, for example [a](mention:a)

```cpp
// !mention one
int factorial(int n) {
  // !mention two
  if (n == 0) {
    // !mention three
    return 1;
    // !mention a
  } else {
    return n * factorial(n - 1);
  }
}
```
</WithMentions>

````

Is it possible to use `!mention` to highligh specific word?
For example with mark, `!mark[/lorem == null/]`

````mdx


<WithMentions>
- Mention specific words [one](mention:one)

```cpp
// !mention[/factorial/] one
int factorial(int n) {
  if (n == 0)
    return 1;
  } else {
    return n * factorial(n - 1);
  }
}
```
</WithMentions>

````

## Tooltip

How to escape brackets to use with tooltip `[]`?  For example `#[program]` throws error:
`[ Server ] Error: Invalid RegExp string: [/#[program]`

````mdx


<WithNotes>

```rs
// !tooltip[/#[program]/] example
#[program]
pub mod cookbook {

}
```

### !example

Not working with #[program]

</WithNotes>

````

It seems tooltips currently are unviewable on mobile since there's no way to hover. Can tooltip annotation be updated to display on tap for mobile?

## Mark

Is it possible to have `!mark` annotation work with `terminal` or `shell`

````mdx

```terminal
$ deploy
// !mark
Deployment successful. Completed in 17s.
```

```shell
$ deploy
// !mark
Deployment successful. Completed in 17s.
```

````

## CodeTabs

When using `storage` with CodeTab, if there isn't a matching snippet with the same key, none get rendered.
Can it default to first, if one is not found?

```
<CodeTabs storage="cookbook">
```

https://github.com/user-attachments/assets/f90a941f-5c1f-4bad-b650-5c92c9424ea8

## RPC

The the page length on `/getaccountinfo` doesn't  expand to account for the added height from "Try It"  response. It doesn't seem to affect any of the other rpc pages though. https://solana.com/docs/rpc/http/getaccountinfo

ZYJLiu:
@pomber any suggestions for the examples listed above?

pomber:
[...]

PR #499, opened 2025-05-10
Better mention annotation

pomber:
Related #288

### Summary of Changes

- Arbitrary labels for the `!mention` annotation
- Allow inline usage to highlight specific words
- Added background color to highlighted mentioned lines to make it more consistent with the inline highlight


ZYJLiu:
thank you!

ZYJLiu (approved):

PR #507, opened 2025-05-13
Improve tooltip

pomber:
### Summary of Changes

- Remove the extra bottom margin in tooltips
- Better regexp support in annotations. This should work now: (note the escaped `[]` since it's a regexp)

````
```rs
// !tooltip[/#\[program\]/] example
#[program]
pub mod cookbook {
```
````

Since it's a regexp, you can even use `.` as wildcard to match any char:

````
```rs
// !tooltip[/..program./] example
#[program]
pub mod cookbook {
```
````

Related #288

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "cheerio": "^1.0.0",
     "class-variance-authority": "^0.7.1",
     "classnames": "^2.5.1",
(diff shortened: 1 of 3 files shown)


ZYJLiu:
thank you!

ZYJLiu (approved):

PR #510, opened 2025-05-14
Fix selected tab when stored title doesn't exist

pomber:
### Problem

When using `storage` with CodeTab, if there isn't a matching snippet with the same key, none get rendered.

### Summary of Changes

- Default to first tab if the stored tab title no longer exists.

Related #288


ZYJLiu (approved):

ZYJLiu:
thank you

PR #549, opened 2025-06-10
fix: update link to correct secp256r1 program docs

metasal1:
Current incorrect link: https://docs.rs/solana-secp256k1-recover/latest/solana_secp256k1_recover/index.html

Correct link: https://docs.rs/solana-secp256r1-program/latest/solana_secp256r1_program/all.html

### Problem

The current link to the secp256r1 instructions points to the k1 program

### Summary of Changes

Updated link to the correct address

Fixes #

Point to R1 instead of K1


h4rkl (approved):
LGTM :)

PR #564, opened 2025-06-19
Fix incorrect wording about bump value

0xjuaan:
### Problem
The doc currently states:

> If `find_program_address` has to take a long time to find a valid address,
meaning it has a high bump

If `find_program_address` takes a long time to find a valid address, this would mean that the bump has a low value (rather than a high one). 

This is because `find_program_address` starts at 255 and decrements it each time- so a lower bump would require more iterations. 

This can be seen in the Solana SDK here: https://github.com/anza-xyz/solana-sdk/blob/39c3ffd0fd7cafe1874d211e0ba8c6e2f75eac54/pubkey/src/lib.rs#L828-L840 

### Summary of Changes
Changed it from 'low bump' to 'high bump'

Fixes #

### DIFF
--- a/content/guides/advanced/how-to-optimize-compute.mdx
+++ b/content/guides/advanced/how-to-optimize-compute.mdx
@@ -178,7 +178,7 @@ but it's important to be aware of the compute usage of `find_program_address`
 and how you can optimize it.
 
 If `find_program_address` has to take a long time to find a valid address,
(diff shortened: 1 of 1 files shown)


ZYJLiu:
thank you!

ZYJLiu (approved):

PR #578, opened 2025-06-27
feat: setup jest tests

h4rkl:
### Problem

Currently there is zero coverage for integration testing in the repo. This piece of work aims to front-load some basic tests for i18n in preparation for fixing the i18n bugs currently live in prod. 

### i18n bug background

The issue with i18n in prod looks to be conflicting libs shared between fumadocs and the core next app. There's currently discussion going on in the engineering team to get consensus on an approach to fix this, but it doesn't make sense to leave this branch idle in the mean time. It is a distinct piece of work and can be merged in on it's own.

### Testing strategy

There's no official testing strategy for the site currently. As we make changes in the future testing will become an important part of validating the before and after state of the application.

### Summary of Changes

- Added jest libs
- setup jest config for next
- created dedicated __tests__ folder for build exclusion


package.json
```diff
@@ -78,28 +78,33 @@
     "yup": "^1.4.0"
   },
   "devDependencies": {
-    "@babel/core": "^7.25.9",
```
h4rkl, line 82: We don't need babel as jest use native next/jest, babel is included [under the hood of next 15](https://nextjs.org/docs/pages/guides/babel) and we never had a .babelrc file.

I initially setup a .babelrc file for jest and I scrapped it because the next native implementation looks to be pretty well tuned and it made `yarn dev` run very slow.

pomber:
lgtm

PR #585, opened 2025-07-01
fix: clarify getEpochSchedule docs

t-nelson:
while a cluster's epoch schedule is _originally_ specified in genesis. it is in fact stored in out-of-band (not accounts) cluster state and can be changed via feature gate in the future


ZYJLiu:
thank you!

PR #605, opened 2025-07-13
fix(intro-to-reading-data.md): set `module` to `Node16` to match `moduleResolution: Node16`

dimmasyusuf:
### Problem

Option 'module' must be set to 'Node16' when option 'moduleResolution' is set to 'Node16'.ts

Specify what module code is generated.
See more: https://www.typescriptlang.org/tsconfig#module

### Summary of Changes

set `module` to `Node16` to match `moduleResolution: Node16`


h4rkl:
Node 16 is outdated and not correct here sorry. 

See Node release schedule for ref - https://nodejs.org/en/about/previous-releases

PR #638, opened 2025-07-28
docs(tokens): replace token-22 instances with og token program

jim4067:
### Summary of Changes

Replace instances of token-2022 in tokens section to use original SPl token program


ZYJLiu:
thank you!

PR #646, opened 2025-07-29
Add Inkeep Chat to entire site

ZYJLiu:
- add Inkeep AI Chat widget to entire solana-com


ZYJLiu:
bumping `@inkeep/cxkit-react` seems to have cause the preview builds to fail. 
reverted to just adding `InkeepChatButton` component on all app/page router pages

PR #713, opened 2025-08-19
Update GH action from Node v18 -> v24

BriungRi:
### Problem

CI running on old Node v18

### Summary of Changes

Update to v24

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Use Node.js node
         uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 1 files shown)


GuiBibeau (approved):

.github/workflows/action.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Use Node.js node
         uses: actions/setup-node@v2
         with:
-          node-version: "18"
+          node-version: "22"
```
beeman: Might as well update it to v24 while you're at it as v22 will be EOL in a few months.

PR #737, opened 2025-09-09
chore: switch to pnpm

beeman:
### Problem

This repo is currently using `yarn@1.22.22` which is getting pretty outdated. Most other modern projects in this organization use `pnpm`. 
 

### Summary of Changes

- [x] Switch to a pnpm workspace.
- [x] Remove yarn.lock
- [x] Switch CI to use pnpm


.github/workflows/action.yml
```diff
@@ -11,30 +11,38 @@ jobs:
...
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10
+          run_install: false
+      - name: Install Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: "18"
-      - name: before_install
+          node-version: "24"
```
h4rkl: Small thing but Vercel uses v22 (their current highest -v) here so can we align versions for consistency pls. Also for any other node -v.

beeman: Updated!

PR #743, opened 2025-09-10
fix: ensure we only load supported locales

beeman:
### Problem

The dev server logs show this error message when you visit the site with the dev tools open and Chrome tries to download
`.well-known/appspecific/com.chrome.devtools.json`:

```
Error: Cannot find module './.well-known/common.json'
    at <unknown> (.next/server/pages/[locale]/[...slug].js:249:12)
    at async getStaticProps$1 (webpack-internal:/(pages-dir-node)/src/pages/[locale]/src/pages/[locale]/[...slug].js:91:23)
  89 |     const page = await getPage(slug, locale);
  90 |     console.log("locale", locale);
> 91 |     const messages = (await import(`@@/public/locales/${locale}/common.json`))
     |                       ^
  92 |       .default;
  93 |
  94 |     return { {
  code: 'MODULE_NOT_FOUND'
```

### Summary of Changes

Check if the `locale` we receive exists in the `locales` array. If not, treat it as not found.


h4rkl:
If you switch back to this repo as `git remote` and use branches off`main` @beeman these previews will build.

PR #756, opened 2025-09-13
debug: replace fs-mdx (WIP)

h4rkl:
### Problem

Recreating PR #750 after revert (#755) to continue work on fs-mdx for @pomber 

Have a snapshot of the error logs for you in Slack @pomber. If we can create tests that address the failures and then deploy any fixes that would be ideal.


h4rkl:
Will close this for now and it can be reopened if you plan to work on it further @pomber  🤝

PR #780, opened 2025-09-24
feat: add generator to create monorepo packages

beeman:
### Problem

We need a way to quickly scaffold out new packages in the `packages/` directory in a consistent way.

### Summary of Changes

Use [turbo gen](https://turborepo.com/docs/guides/generating-code) to create new packages.

You can run the generator in interactive mode:

```shell
pnpm turbo gen
```

Or you can run pass the parameters and automate the generation:

```shell
pnpm turbo gen pkg --args basic i18n
#              ^          ^     ^                           
#              |          |     |
#              |          |     + - - the name of the package
#              |          |    
#              |          + - - the type of package you want to generate                           
#              |             
#              + - - the generator you want to invoke                           
```

Fixes #


turbo/generators/generator-pkg.ts
```diff
@@ -0,0 +1,34 @@
+import { PlopTypes } from "@turbo/gen";
+import { validateName } from "./helpers";
+
+const TARGET_DIR = "packages";
+const TEMPLATE_ROOT = "templates/pkg";
+
+export const generatorPkg: PlopTypes.PlopGeneratorConfig = {
+  description: `Generator a package in the ${TARGET_DIR} directory`,
```
h4rkl: typo

turbo/generators/generator-pkg.ts
```diff
@@ -0,0 +1,34 @@
...
+      type: "list",
+      name: "type",
+      message: "What type of package should be created?",
+      choices: ["basic", "react-ui"],
+    },
+    {
+      type: "input",
+      name: "name",
+      default: ({ type }) => type,
+      message: "What is the name of package?",
```
h4rkl: "What is the name of the package?"

turbo/generators/templates/pkg/basic/package.json.hbs
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@workspace/{{name}}",
+  "version": "0.0.0",
+  "private": false,
+  "license": "MIT",
```
h4rkl: Can we standardise licences across the app - should match root `"license": "GPL-3.0-or-later",`.

Also need to update `/solana-com/packages/config-typescript/package.json` as currently showing PROPRIETARY

PR #783, opened 2025-09-24
fix: Grammar mistake in accouts.mdx

Wilson-Wu1:
### Problem
Small grammar mistake in the Accounts doc: 
_"For non-executable accounts, this generally stores state that's meant be read from."_

Missing "to" after "meant"

### Summary of Changes

Added missing "to"


ZYJLiu:
thank you!

PR #785, opened 2025-09-25
chore: move i18n configuration to separate package

beeman:
### Problem

In order to have UI components depend on the i18n configuration, it must live in a package instead of `apps/web`.

### Summary of Changes

- Create a package for i18n 
- Update references to import from the package
- Add `ts-loader` to the Next webpack config so that `getAllUrls` from `getUrls.js` can read it
- I had to update a few dependencies (next + eslint) to make sure I didn't get a `JavaScript heap out of memory` error

image

Fixes #


apps/web/package.json
```diff
@@ -42,7 +43,7 @@
     "lodash": "^4.17.21",
     "lucide-react": "^0.475.0",
     "mermaid": "^11.5.0",
-    "next": "^15.3.5",
+    "next": "^15.5.4",
```
h4rkl, line 46: Fuck yeah

PR #786, opened 2025-09-25
Use `fumadocs-core` for loading localized folders

pomber:
### Problem

We used `fs-mdx` to load content from folders for each locale, but now it's [supported in `fumadocs-core@15.6.10`](https://fumadocs.dev/docs/headless/page-conventions#i18n-routing).

### Summary of Changes

- Update `fumadocs-core` to v15.6.10
- Use `parser: "dir"` for localized sources
- Fix `remarkImage` errors on Vercel Functions
- Remove `lastModifiedDate`, we are not using it anymore and it was slowing down `.source/index.ts` generation


apps/web/package.json
```diff
@@ -48,6 +48,7 @@
     "node-fetch-cache": "^5.0.2",
     "postcss-flexbugs-fixes": "^5.0.2",
     "posthog-js": "^1.232.7",
+    "prop-types": "^15.8.1",
```
ZYJLiu, line 51: is this dependency being used?

pomber: I was getting an error that "prop-types" was missing, but that was even before changing anything I think, not sure why it wasnt failing on CI

apps/web/package.json
```diff
@@ -36,7 +36,7 @@
     "date-fns-tz": "^3.2.0",
     "feed": "^4.2.2",
     "fs-mdx": "0.4.0",
-    "fumadocs-core": "15.0.3",
+    "fumadocs-core": "15.6.10",
     "fumadocs-ui": "14.7.0",
```
h4rkl, line 40: What are the risks of using mismatched versions here? I see we've ignored types in tree above but do we understand the downstream effects in the app for this mismatch? Not against it in principle if we're incrementally migrating to 15 but we need to be 100 that this isn't going to create edge cases that blow out the docs.

pomber: Risk from the mismatch should be low. `fumadocs-ui` consumes `fumadocs-core`, so a lower `ui` with a higher `core` is safe, especially since we’re keeping `core` in the same major (15.0.3 → 15.6.10). The plan is to migrate from `fs-mdx` to the latest `fumadocs-mdx` in the next PR, and once we have Tailwind 4, move to the latest `fumadocs-ui`.

h4rkl: Keep in mind that once we've got the header/footer package we'll be able to split the docs out into its own app in the turbo repo. Not sure if this changes your delivery path but probably good to be aware of.

ZYJLiu:
@pomber, had to rollback the vercel deployment. I had only checked the `/docs` pages and locales, but it looks like the `/courses` pages are not working.

PR #788, opened 2025-09-25
round 1 edits to the quick start guide

lrobledo:
### Problem

I tested this guide as a new user and got set up for the first time on the Solana Playground.
Edits are to lower readability and to add clarification as needed.

### DIFF
--- a/apps/web/content/docs/en/intro/quick-start/index.mdx
+++ b/apps/web/content/docs/en/intro/quick-start/index.mdx
@@ -124,4 +124,4 @@ Solana's [Web Faucet](https://faucet.solana.com/).
 ## Next Steps
 
 Now that you have created a Playground wallet and added Devnet SOL, you are ready to
-[deploy your first program](/docs/intro/quick-start/deploying-programs).
+[read from the Solana network](/docs/intro/quick-start/reading-from-network).


apps/web/content/docs/en/intro/quick-start/index.mdx
```diff
@@ -1,124 +1,130 @@
...
-The system prompts you to save your wallet's keypair. Once ready, click
-"Continue" to proceed.
+[//]: # (will additional info help me decide where I want to save this? )
 
 ![Create Playground Wallet](/assets/docs/intro/quickstart/pg-create-wallet.png)
 
-You should now see your wallet's address, SOL balance, and connected cluster
-(devnet by default) at the bottom of the window.
+At the bottom of the screen, you will see that you are now connected to the Playground wallet.
+You will also see the [connected cluster](/docs/references/terminology#connected-cluster) (`devnet`),
```
lrobledo, line 68: lmk if you think these lines are carrying their weight. I linked to the terms page, and removed in the in-line definitions on lines 75-88

apps/web/content/docs/en/intro/quick-start/index.mdx
```diff
@@ -1,124 +1,130 @@
...
+<Step>
+
+### Save your Keypair
+
+Click **Save keypair** to save your key pair locally.
+When you're ready, click **Continue**.
 
-The system prompts you to save your wallet's keypair. Once ready, click
-"Continue" to proceed.
+[//]: # (will additional info help me decide where I want to save this? )
```
lrobledo: for example, how secure should it be?

apps/web/content/docs/en/intro/quick-start/index.mdx
ZYJLiu: linking to terms seems fine

apps/web/content/docs/en/intro/quick-start/index.mdx
ZYJLiu: You only need to save the keypair generated by Solana Playground if you want to recover the account or use it outside of Solpg. It doesn't matter where the user saves it.

Generally unnecessary because Solpg should only be used for learning and it is not recommended to send mainnet assets to the generated address (since the keypair is also in locale storage in the browser).

ZYJLiu:
not sure why the vercel preview deployment failed, does it run for you locally?

lrobledo:
> not sure why the vercel preview deployment failed, does it run for you locally?

No, it doesn't. Looking at it now

apps/web/content/docs/en/intro/quick-start/index.mdx
```diff
@@ -4,121 +4,124 @@ description:
...
 
 ![Faucet Airdrop](/assets/docs/intro/quickstart/faucet-airdrop.gif)
 
 </Step>
 </Steps>
+
+## Next Steps
+
+Now that you have created a Playground wallet and added Devnet SOL, you are ready to
+[deploy your first program](/docs/intro/quick-start/deploying-programs).
```
ZYJLiu: The next page is `/reading-from-network`, but this recommends going straight to `/deploying-programs`.

How do you suggest revising the flow of this?

The intent is that this first page sets up Solpg, but the readers should first go through reading/writing to network pages before deploying first program.

Should there be a note recommending reading/writing first, with option to skip straight to deploying or something else like reordering the pages?

lrobledo: Good point. I changed it to Reading from the Network, if we need to revisit it later on then we can.

lrobledo:
@ZYJLiu When I checked this morning, the Vercel deploy had worked (4 days ago) but then I pushed more changes and its failing again. I'm not sure what changed there

ZYJLiu:
thank you

PR #790, opened 2025-09-25
round 1 edits to the deploy a program guide

lrobledo:
### Summary of Changes

Revisions of the Deploy Program guide.

### DIFF
--- a/apps/web/content/docs/en/intro/quick-start/deploying-programs.mdx
+++ b/apps/web/content/docs/en/intro/quick-start/deploying-programs.mdx
@@ -1,45 +1,40 @@
 ---
-title: Deploying Programs
+title: Deploy a Program
 description:
   Learn how to build, deploy, and test your first Solana program using the
   Anchor framework and Solana Playground. This beginner-friendly guide walks
   through creating a simple program, deploying it to devnet, running tests, and
   closing the program.
-h1: Deploying Your First Solana Program
+h1: Deploy Your First Solana Program
 ---
 
-In this section, you'll build, deploy, and test a simple Solana program (smart
-contract) using the Anchor framework. By the end, you'll have deployed your
-first program to the Solana blockchain.
-
-The purpose of this section focuses on familiarizing you with the Solana
-Playground. The guide walks through a more detailed example in the Program
-Derived Address (PDA) and Cross-Program Invocation (CPI) sections. For more
-details, refer to the [Programs on Solana](/docs/core/programs) page.
+Build, deploy, and test your first [program](/docs/core/programs)
+to the Solana Blockchain using the Anchor framework.
+This guide will help you become more familiar with the Solana Playground
+and prepare you to use program derived addresses (PDA)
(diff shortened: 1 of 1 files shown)


apps/web/content/docs/en/intro/quick-start/deploying-programs.mdx
```diff
@@ -404,54 +386,45 @@ Transaction executed in slot 308150984:
...
+You built and deployed your first Solana program using Anchor.
 
 </Step>
 </Steps>
+
+## Next steps
+
+Now that you have deployed your first program, you are ready to explore
+[cross-program invocation](/docs/intro/quick-start/cross-program-invocation)
+and [program-derived addresses](/docs/intro/quick-start/program-derived-address).
```
ZYJLiu: minor note, users should read the program derived address page before cross program invocation page.

The reason is because pdas can be used when making a cpi, so readers should learn about pdas first.

apps/web/content/docs/en/intro/quick-start/deploying-programs.mdx
```diff
@@ -97,74 +92,65 @@ pub struct NewAccount {
...
-account.
+The number of bytes to allocate for the account
 
 </WithNotes>
 
 <Accordions>
-<Accordion title="Explanation">
 
-For now, this covers the high-level overview of the program code:
+[//]: # (Is this redundant?)
```
ZYJLiu: Yea, it is a bit redundant. 

The reason for including is because the on-hover tooltip doesn't work in mobile view.
It also provides slightly more context than the tooltip.

apps/web/content/docs/en/intro/quick-start/deploying-programs.mdx
```diff
@@ -1,45 +1,40 @@
...
-This basic Solana program creates a new account and stores a number in it. The
-program contains one instruction (`initialize`) which:
-
-- Requires a _rs`data: u64`_ parameter as input
+- Receives a required parameter called _rs`data: u64`_
 - Creates a new account
-- Saves the value of the _rs`data: u64`_ parameter in the account's data
-- Logs a message to the transaction's program logs
+- Saves the value of _rs`data: u64`_ to the new account
+- Saves a message to the program logs
```
ZYJLiu: > Saves a message to the program logs

The logs aren't really "saved", its printed to program logs.
Not sure if "saves" is the correct word here.

For example, when data is saved to an account, the data is stored on the account and can be fetched using the account's address.

apps/web/content/docs/en/intro/quick-start/deploying-programs.mdx
lrobledo: makes sense, thanks for the info

apps/web/content/docs/en/intro/quick-start/deploying-programs.mdx
lrobledo: How does 'adds' sound?

PR #796, opened 2025-09-27
Use `fumadocs-core` for loading localized folders

pomber:
### Problem

We used `fs-mdx` to load content from folders for each locale, but now it's [supported in `fumadocs-core@15.6.10`](https://fumadocs.dev/docs/headless/page-conventions#i18n-routing).

### Summary of Changes

- Update `fumadocs-core` to v15.6.10
- Use `parser: "dir"` for localized sources
- Fix `remarkImage` errors on Vercel Functions
- Remove `lastModifiedDate`, we are not using it anymore and it was slowing down `.source/index.ts` generation


pomber:
Moved to #798 to unstuck vercel deployment

PR #810, opened 2025-10-01
Fix scrollycoding `rootMargin`

pomber:
### Problem

The intersection observer from the ScrollyCoding component wasn't triggering when the browser height wasn't large enough.

### Summary of Changes

Give more space to the `rootMargin`


ZYJLiu (approved):

ZYJLiu:
@pomber, unrelate to this PR but just noticed your slack account got deactived.
Trying to figure out why and resolve.

PR #842, opened 2025-10-10
feat: add templates site content

beeman:
### Problem

The template site lives in a separate repo

### Summary of Changes

Add the content to the base template site

Follow up of #841

### DIFF
--- a/apps/templates/components/app-hero.tsx
+++ b/apps/templates/components/app-hero.tsx
@@ -15,14 +15,14 @@ export function AppHero({
       <div className="absolute top-0 left-0 w-full h-full -z-50 overflow-x-hidden">
         <Image
           src={"/hero.webp"}
-          alt="Hero"
+          alt=""
           width={1000}
           height={1000}
           className="absolute top-0 -left-[550px] hidden md:block -z-10"
         />
         <Image
           src={"/hero.webp"}
-          alt="Hero"
+          alt=""
           width={1000}
           height={1000}
           className="absolute -top-44 -right-[550px] hidden md:block -z-10"
--- a/apps/templates/components/hooks/use-search-keyboard.ts
+++ b/apps/templates/components/hooks/use-search-keyboard.ts
@@ -1,5 +1,6 @@
 import { useState, useCallback } from "react";
 import { RepokitTemplate } from "@/lib/generated/repokit";
+import { useRouter } from "next/navigation";
 
 export interface UseSearchKeyboardReturn {
   activeIndex: number;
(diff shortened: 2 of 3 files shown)


apps/templates/components/app-hero.tsx
```diff
@@ -0,0 +1,50 @@
...
+  children,
+  subtitle,
+  title,
+}: {
+  children?: React.ReactNode;
+  subtitle?: React.ReactNode;
+  title?: React.ReactNode;
+}) {
+  return (
+    <div className="flex flex-col items-center justify-center py-[16px] md:py-32">
```
h4rkl, line 14: parent needs `relative` positioning for `absolute` child

apps/templates/components/app-hero.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <div className="absolute top-0 left-0 w-full h-full -z-50 overflow-x-hidden">
+        <Image
+          src={"/hero.webp"}
+          alt="Hero"
```
h4rkl: Use empty alt text for decorative images - same on L25

apps/templates/components/hooks/use-search-keyboard.ts
```diff
@@ -0,0 +1,86 @@
...
+          setActiveIndex((prev) =>
+            prev > 0 ? prev - 1 : filteredTemplates.length - 1,
+          );
+          break;
+        case "Enter":
+          e.preventDefault();
+          if (activeIndex >= 0 && filteredTemplates[activeIndex]) {
+            const template = filteredTemplates[activeIndex];
+            const internalHref = `/developers/templates/${template.source.id}/${template.name}`;
+            window.location.href = internalHref;
```
h4rkl: Better to use app router `router.push` here to preserve state and avoid full page reload

apps/templates/package.json
h4rkl, line 2: If we can update this to "templates"

apps/templates/components/app-hero.tsx
beeman: It seems like setting `relative` here makes the hero image contain to the parent element (`AppHero`) which in this case is not what we want. 

image

image

apps/templates/components/app-hero.tsx
beeman: TIL! Fixed!

GuiBibeau:
@beeman I'm moving a lot of stuff around templates to review new dev onboarding, are you ok if i take this over?

beeman:
> @beeman I'm moving a lot of stuff around templates to review new dev onboarding, are you ok if i take this over?

@GuiBibeau go for it ser!

PR #848, opened 2025-10-15
fix: prevent infinite reload when multiple locale segments exist in URL

### LINKED ISSUE #847: Small bug in handling multiple consecutive locales in a url that causes the page to reload infinitely
When you go to https://solana.com/ja/us/docs or any url of form https://solana.com/\<locale\>/\<locale\>/docs it triggers infinite reloading of the page.

sivajisj:
### Problem

Visiting URLs of the form https://solana.com/<locale>/<locale>/docs (for example, https://solana.com/ja/ja/docs) causes an infinite redirect loop due to duplicate locale segments being added by the locale handling middleware.

### Summary of Changes
Added a check in the locale middleware to prevent re-adding a locale if it’s already present in the URL path.

Updated apps/website/middleware.ts to handle this edge case cleanly.

Modified file:
apps/website/middleware.ts

Fixes #847

### DIFF
--- a/apps/web/src/middleware.ts
+++ b/apps/web/src/middleware.ts
@@ -6,15 +6,32 @@ import { locales } from "@workspace/i18n/config";
 const handleI18nRouting = createMiddleware(routing);
 
 export default async function middleware(req: NextRequest) {
+  const pathname = req.nextUrl.pathname;
+
   // Skip i18n for /breakpoint/* paths
-  if (req.nextUrl.pathname.startsWith("/breakpoint")) {
+  if (pathname.startsWith("/breakpoint")) {
     return NextResponse.next();
   }
 
-  if (req.nextUrl.pathname !== req.nextUrl.pathname.toLowerCase()) {
+  if (pathname !== pathname.toLowerCase()) {
     return NextResponse.redirect(
-      `${req.nextUrl.origin + req.nextUrl.pathname.toLowerCase()}`,
+      `${req.nextUrl.origin + pathname.toLowerCase()}`,
     );
+  }
+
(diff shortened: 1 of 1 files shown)


apps/web/src/middleware.ts
```diff
@@ -6,17 +6,34 @@ import { locales } from "@workspace/i18n/config";
...
+  const pathname = req.nextUrl.pathname;
+
   // Skip i18n for /breakpoint/* paths
-  if (req.nextUrl.pathname.startsWith("/breakpoint")) {
+  if (pathname.startsWith("/breakpoint")) {
     return NextResponse.next();
   }
 
-  if (req.nextUrl.pathname !== req.nextUrl.pathname.toLowerCase()) {
+  if (pathname !== pathname.toLowerCase()) {
```
h4rkl: The string concatenation approach constructs only the origin and pathname, discarding any query parameters or hash fragments from the original URL. For example, /JA/docs?search=test#section would redirect to /ja/docs, losing ?search=test#section.

Apply this diff to preserve all URL components:

-  if (pathname !== pathname.toLowerCase()) {
-    return NextResponse.redirect(
-      `${req.nextUrl.origin + pathname.toLowerCase()}`,
-    );
-  }
+  if (pathname !== pathname.toLowerCase()) {
+    const url = new URL(req.url);
+    url.pathname = pathname.toLowerCase();
+    return NextResponse.redirect(url);
+  }

h4rkl:
Appreciate this PR but there's a cleaner way of doing this by using locales config and it was faster to just push the fix in 637e5e71c159e4756875673683978d48437165f0 rather than go through code review here.

PR #861, opened 2025-10-17
Extract header/footer to new package

pomber:
### Problem

The header/footer elements need to be shared between apps

### Summary of Changes

- Create a new `ui-chrome` package
- Move header, footer, and their dependencies to the new package

FYI:
- This PR focuses mostly on extracting the components as they are, without any improvement to the implementation
- Didn't move other components like `CookieConsent` and `SitewideTopAlert` because I wasn't sure if it's needed


apps/web/src/components/breakpoint/BreakpointHeader.jsx
```diff
@@ -1,5 +1,5 @@
 import Navbar from "react-bootstrap/Navbar";
-import Link from "../shared/Link";
+import Link from "../../utils/Link";
```
h4rkl: Might as well make this @/utils also

h4rkl:
Approved, but did you a favour @pomber and fixed thee merge conflicts in #868 before pushing to prod

PR #864, opened 2025-10-18
Fix: replace incorrect '\\n' with proper newline '\n' in logs

sivajisj:
Replaced '\\n' with '\n' in all console.log statements to ensure proper line breaks in terminal output.

### Problem
The console.log statements in createAndMintToken.ts were using \\n, which prints a literal backslash and n instead of creating a newline. This makes logs harder to read during token creation and minting.

### Summary of Changes
Replaced all occurrences of \\n with \n in console.log statements.
Ensures proper line breaks in terminal output for better readability.
Fixes # (You can put the relevant issue number here if you have one)

Fixes #


sivajisj:
Thank you for accepting : ),
Glad to contribute to this project

PR #920, opened 2025-11-01
Add Pylint workflow for Python code analysis

elkinsjustin1218-cmyk:
### Summary of Changes

Fixes #


h4rkl:
Appreciate this PR but we only have a single .py file and there's no plans to add more at this stage so a workflow linter is out of scope.

PR #935, opened 2025-11-07
fix: derive ata instead of creating a keypair for cpi and default state examples

datasalaryman:
Solana Docs recommend generating a keypair when creating a token account. This isn't standard practice and should be avoided. 

All references to generating a keypair for the token account have been replaced with finding the token account with seeds.

### DIFF
--- a/apps/web/content/docs/en/tokens/basics/create-token-account.mdx
+++ b/apps/web/content/docs/en/tokens/basics/create-token-account.mdx
@@ -54,7 +54,7 @@ account tracks:
 
 There are two approaches to creating token accounts:
 
-1. Create the account from scratch (we use a keypair in the example below)
+1. Create the account using the public key of a keypair as the address
 2. Create an associated token account (ATA) with a deterministic address
 
 ### Example: USD Coin (USDC)


datasalaryman:
I missed the token 2022 references. Will work on them tomorrow

datasalaryman:
this is ready for review. thanks

ZYJLiu:
the snippets currently do not run on the vercel preview deployment with the changes. 
can you check locally that the snippets work. you can use the existing snippets on solana-com as reference

ZYJLiu:
> @ZYJLiu - feel free to check now. I'm having a hard time testing it locally. Let me know if you have a way that works.

are you updating these pages without checking the updated snippets work?

datasalaryman:
hi @ZYJLiu I'm unable to on local because it takes awhile to run a dev server for this project on localhost. I also don't have access to the vercel project. 

I have a pesonal project I'm basing these examples on, but it isn't clear if it's translating one-for-one. 

Is there a way perhaps I can run a `--filter` flag when running the docs? or is there a better way?

PR #1038, opened 2026-01-10
chore(privacyhack): New bounties, workshops, resources

catmcgee:
### Summary of Changes

Fixes #

### DIFF
--- a/apps/web/public/locales/en/common.json
+++ b/apps/web/public/locales/en/common.json
@@ -3810,8 +3810,8 @@
         {
           "sponsor": "MagicBlock",
           "logo": "",
-          "title": "Real-time Privacy",
-          "description": "Build real-time, fully confidential applications using MagicBlock's Private Ephemeral Rollups on Solana. Combine confidential state, real-time execution, and compliance-first approach.\n\nBest Private App: $3k\nSecond Place: $1.5k\nThird Place: $500",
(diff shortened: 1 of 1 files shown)


apps/web/public/locales/en/common.json
```diff
@@ -3715,6 +3715,13 @@
           "speaker": "Encrypt.trade Team",
           "speakerTitle": "Encrypt.trade",
           "url": "https://x.com/solana_devs"
+        },
+        {
+          "title": "Arcium for Anchor Developers",
+          "date": "TBD",
+          "speaker": "QuickNode Team",
```
GuiBibeau: Is that right? Title says Arcium but speaker says quicknode

apps/web/public/locales/en/common.json
```diff
@@ -3715,6 +3715,13 @@
...
+          "speakerTitle": "QuickNode",
+          "url": "https://x.com/solana_devs"
```
GuiBibeau, line 3724: is that the right url?

apps/web/public/locales/en/common.json
catmcgee: yep!

apps/web/public/locales/en/common.json
catmcgee: workshops will be streamed there so yes just linking to the X profile

PR #1186, opened 2026-02-24
feat: Prediction Markets Hack page

catmcgee:
Adds Prediction Market hackathon page with no bounties yet so companies can add bounties


apps/web/src/app/[locale]/prediction-markets-hack/prediction-markets-hack.tsx
```diff
@@ -0,0 +1,3031 @@
...
+          />
+        </svg>
+      </div>
+
+      {/* Hero Section */}
+      <section className="relative min-h-screen flex items-center justify-center overflow-hidden">
+        {/* Market animation keyframes */}
+        <style
+          dangerouslySetInnerHTML={{
+            __html: `
```
Comment (GuiBibeau) [inline apps/web/src/app/[locale]/prediction-markets-hack/prediction-markets-hack.tsx]: nit: Might make sense to move that into a component in the same folder called animation.tsx or something.

PR #1203, opened 2026-03-01
fix(accelerate): add favicon.svg to prevent i18n locale routing error

### LINKED ISSUE #1202: Bug: Invalid language tag "favicon.svg" in accelerate app causes i18n errors
## Steps To Reproduce
1. Run `pnpm dev` from the repo root
2. Open the accelerate app at http://localhost:3004
3. Load the page (or request /favicon.svg)

## The current behavior
Console shows repeated errors: `Error: INVALID_MESSAGE: Invalid language tag: favicon.svg` in `Speakers.tsx` and `Sponsors.tsx`. The i18n `t()` function receives "favicon.svg" as the locale instead of a valid locale like "en".

This happens because the accelerate app has no `public/favicon.svg`. When the browser requests `/favicon.svg`, the middleware excludes it (paths with dots), so Next.js matches it to the `[locale]` dynamic route. The first path segment "favicon.svg" becomes the locale param.

Image

## The expected behavior
The favicon should load without errors. Requests to `/favicon.svg` should be served as a static file, not routed through the `[locale]` segment.

Didier-ds:
### Problem
The accelerate app throws Invalid language tag: favicon.svg i18n errors when loading. The browser requests /favicon.svg, but the app has no public/favicon.svg. Because the middleware skips paths with dots, the request hits the [locale] route and favicon.svg is treated as the locale, which breaks the translation function.

### Summary of Changes
Add public/favicon.svg to the accelerate app (same Solana favicon as the main web app)
Ensures /favicon.svg is served as a static file instead of being matched by the [locale] route
Aligns with config.siteIcon and metadata in metadata.ts

Fixes #1202


h4rkl:
Closing this as we're about to migrate shared elements like this into ui-chrome

PR #1222, opened 2026-03-11
feat: add /skills page

MicaiahReid:


apps/web/src/app/[locale]/skills/skills.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+      skills = results
+        .filter(
+          (r): r is PromiseFulfilledResult<SkillItem> =>
+            r.status === "fulfilled",
+        )
+        .map((r) => r.value);
+    }
+  } catch {
+    // Render gracefully with empty skills on error
```
Comment (davidmurdoch) [inline apps/web/src/app/[locale]/skills/skills.tsx]: Have you considered retry with exponential backoff?

apps/web/src/app/[locale]/skills/skills.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { getTranslations } from "next-intl/server";
+import { SkillItem } from "@/components/skills/SkillCard";
+import { SkillsGrid } from "@/components/skills/SkillsGrid";
+import { SkillsInstallCommand } from "@/components/skills/SkillsInstallCommand";
+
+function parseSkillMarkdown(filename: string, content: string): SkillItem {
+  const slug = filename.replace(/\.md$/, "");
+  const { data } = matter(content);
+
+  return {
```
Comment (davidmurdoch) [inline apps/web/src/app/[locale]/skills/skills.tsx:20]: if you can "compute" the skills category here, then you don't need the double category computation in `SkillCard.tsx` later. You can go further and set the label (from `labelKey`), simplifying the other component logic a little bit more.

apps/web/src/app/[locale]/skills/skills.tsx
```diff
@@ -0,0 +1,116 @@
...
+    slug,
+    title: data.title ?? slug,
+    description: data.description ?? "",
+    githubUrl: `https://github.com/solana-foundation/solana-dev-skill/blob/main/skill/references/${filename}`,
```
Comment (davidmurdoch) [inline apps/web/src/app/[locale]/skills/skills.tsx]: the entries in the array returned by https://api.github.com/repos/solana-foundation/solana-dev-skill/contents/skill/references already includes the key `html_url`, which seems to match this URL perfectly. Would it be better to use it here instead?

apps/web/src/app/[locale]/skills/skills.tsx
```diff
@@ -0,0 +1,116 @@
...
+              f.name.endsWith(".md") && f.name !== "SKILL.md",
+          )
+        : [];
+
+      const results = await Promise.allSettled(
+        mdFiles.map(async (file: { name: string; download_url: string }) => {
+          const raw = await fetch(file.download_url, {
+            next: { revalidate: 3600 },
+          });
+          const content = await raw.text();
```
Comment (davidmurdoch) [inline apps/web/src/app/[locale]/skills/skills.tsx]: nit: Github API likes to be flaky, and is prone to rate limiting (note: it is possible there is a fetch wrapper that adds github auth headers to increase rate limits), it might be _slightly_ more useful to check `raw.ok` and then throw a specific exception for the case where it fails, rather then receive an error due to `parseSkillMarkdown` failing to parse the github error page HTML. Additionally, bailing early when `!raw.ok` avoids downloading the error page, so its technically faster (in the worst case).

apps/web/src/app/[locale]/skills/skills.tsx
Comment (davidmurdoch) [inline apps/web/src/app/[locale]/skills/skills.tsx:71]: It could be nice to log the error here?

apps/web/src/components/skills/SkillsInstallCommand.tsx
```diff
@@ -0,0 +1,35 @@
...
+const COMMAND =
+  "npx skills add https://github.com/solana-foundation/solana-dev-skill";
+
+export function SkillsInstallCommand({ copyLabel }: { copyLabel: string }) {
+  const [copied, setCopied] = useState(false);
+
+  async function handleCopy() {
+    await navigator.clipboard.writeText(COMMAND);
+    setCopied(true);
+    setTimeout(() => setCopied(false), 2000);
```
davidmurdoch, line 16: If you click the button multiple times very quickly, things can get a bit, uh, wonky.

Example:
https://github.com/user-attachments/assets/15f3691f-b13d-4fb1-a31e-1bc2381ed682

You can capture the `setTimeout` return value, then call `clearTimeout` before each call to `navigator.clipboard.writeText` to ensure the timers aren't clobbering each other.

h4rkl:
Nice content for a PR! Can you run the FE Design skill over this `npx skills add https://github.com/anthropics/skills --skill frontend-design` and use `apps/web/src/app/[locale]/solutions/tokenization` as the precedent to draw design look and feel for please.

apps/web/src/app/[locale]/skills/skills.tsx
Comment (MicaiahReid) [reply apps/web/src/app/[locale]/skills/skills.tsx]: Good call, thanks!

apps/web/src/app/[locale]/skills/skills.tsx
Comment (MicaiahReid) [reply apps/web/src/app/[locale]/skills/skills.tsx:20]: We'll have different sections for skills later and we'll be able to dedup logic to keep centralized at the SkillCard level, so I'll keep as is

apps/web/src/app/[locale]/skills/skills.tsx
Comment (MicaiahReid) [reply apps/web/src/app/[locale]/skills/skills.tsx:71]: Yep, thanks!

apps/web/src/components/skills/SkillsInstallCommand.tsx
MicaiahReid: 1. Stop clicking button multiple times very quickly

davidmurdoch: image

MicaiahReid:
> Design looking way better. If we can remove all the rounded edges from the cards we'll be good to go.
> 
> Also, we currently have `apps/web/public/SKILL.md` serving at `solana.com/SKILL.md` and some redirect paths in `apps/web/src/middleware.ts`. Given the convention is `SKILL.md` it probably makes sense to remove the old file from the `public` folder and dynamically pull the latest `https://github.com/solana-foundation/solana-dev-skill` at build time (was copied from here originally). Having `solana.com/SKILLS.md` for agents is an anti-pattern. What do you think?

Good call on the SKILL.md move, done!

PR #1328, opened 2026-04-02
feat(docs:rpc): update wss methods with code snippets

jim4067:
### Summary of Changes
update rpc docs with subscription methods code snippets.

### DIFF
--- a/apps/docs/content/docs/en/rpc/websocket/accountsubscribe.mdx
+++ b/apps/docs/content/docs/en/rpc/websocket/accountsubscribe.mdx
@@ -15,7 +15,7 @@ Subscribe to notifications when an account's lamports or data change.
 
 <APIMethod>
 
-```jsonc !!request
+```jsonc !!request curl
 {
   "jsonrpc": "2.0",
   "id": 1,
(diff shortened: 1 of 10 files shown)


apps/docs/content/docs/en/rpc/websocket/accountsubscribe.mdx
```diff
@@ -51,18 +122,14 @@ Optional configuration object.
 
 ##### !! commitment
 
-!type string  
-!values processed confirmed finalized  
-!default finalized
+!type string !values processed confirmed finalized !default finalized
```
ZYJLiu: can you undo the changes to the annotations for all pages? 
its likely changed due to format on save.

the changes will cause pages either fail to render or the annotations to render incorrectly

jim4067: Updating them now. Looks like there is a commit linting step that runs first.

ZYJLiu (approved):

ZYJLiu:
thank you!

PR #1362, opened 2026-04-09
Fix cookie consent persistence across site apps

afurm:
### Problem

The cookie consent flow only persisted consent and updated `builderNoTrack` when `window.gtag` was already available.

That created a privacy bug where an explicit opt-out could fail to persist if GTM had not initialized yet, and Builder tracking could remain enabled until analytics loaded. The `web` and `docs` variants also only restored `builderNoTrack` on mount when the saved consent value was truthy, so a saved opt-out was not applied consistently after reload.

### Summary of Changes

- normalized the consent state to `null | boolean` with an `isLoaded` guard
- persist `cookie_consent` immediately when consent changes, even if `gtag` is not available yet
- update `builderNoTrack` immediately for both opt-in and opt-out paths
- keep the `gtag("consent", "update", ...)` call conditional on `window.gtag` existing
- applied the fix across the `web`, `docs`, `media`, and `templates` cookie consent components

### DIFF
--- a/apps/docs/src/components/CookieConsent/CookieConsent.jsx
+++ b/apps/docs/src/components/CookieConsent/CookieConsent.jsx
@@ -40,38 +40,44 @@ const setLocalStorage = function (key, value) {
 export default function CookieConsent() {
   const t = useTranslations();
 
-  // cookieConsent is blank by default
-  const [cookieConsent, setCookieConsent] = useState("");
+  // cookieConsent is null by default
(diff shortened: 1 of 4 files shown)


h4rkl:
Thanks for sending this PR through. Have migrated cookie consent to packages so it can be properly shared in /apps in PR https://github.com/solana-foundation/solana-com/pull/1368.

PR #1429, opened 2026-05-01
feat(docs): Test JS cookbook code in CI

catmcgee:
### Problem

Cookbook is easy to get out of date because the code examples just sit in `.md` files

### Summary of Changes

This PR does quite a lot

* Rewrites all Kit & legacy web3js examples into a new directory then pulls them into .md files with `@devrelkit/remark-include-code`
* Tests these code samples against surfpool in CI

We will do the Rust/Python in subsequent PRs


amilz (commented):
This is great! Thank you for doing this.
1. I think you can cut CI setup time by removing CLI/Surfpool deps/setup and use the new [Surfpool SDK](https://www.npmjs.com/package/surfpool-sdk). Would change `.github/workflows/docs-examples.yml` and `packages/docs-examples/test/global-setup.ts`.
2. Might be worth adding some additional checks on some of the logs (e.g., signature could use kit's `signature` or `assertIsSignature` function to verify the payload is as expected:

```ts
    const log = vi.spyOn(console, "log").mockImplementation(() => {});
     //..
    expect(signature(loggedSignature)).toBeTruthy();
```
3. might be worth putting this test on a cron that installs latest dependency packages to help us catch dependency changes to flag that we need to update snippets.

catmcgee:
good ideas @amilz have updated
1. uses surfpool-sdk now, it uses dynamic ports so this forwards 8899/8900 to whatever it picks 
2. now it checks the signature with kit and fetches the tx to assert that `meta.err === null`. i think this is probably enough unless you can think of other high prio things we could test?
3. cron is at at `.github/workflows/docs-examples-deps-drift.yml`, runs mondays

amilz (approved):
🔥 LGTM! Pushed a prettier commit for CI: cf937bf36

PR #1432, opened 2026-05-01
WIP feat(docs) docs examples in Rust

catmcgee:
### Summary of Changes

Fixes #

Merge after https://github.com/solana-foundation/solana-com/pull/1429


.github/workflows/docs-examples.yml
```diff
@@ -61,3 +62,55 @@ jobs:
...
+      - name: Cache Solana CLI
+        id: solana-cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.local/share/solana/install
+          key: solana-cli-stable-${{ runner.os }}-v1
+
+      - name: Install Solana CLI
+        if: steps.solana-cache.outputs.cache-hit != 'true'
+        run: sh -c "$(curl -sSfL https://release.anza.xyz/stable/install)"
```
amilz: I think if we have surfpool, we don't need solana CLI?

catmcgee: it's using keygen. but we could do that in rust instead?

ZYJLiu:
getting similar error on the preview as before on cookbook pages

> Application error: a server-side exception has occurred while loading 

[for example](https://solana-com-docs-5uv2lj8tt-solana-foundation.vercel.app/developers/cookbook/wallets/create-keypair)

.github/workflows/docs-examples.yml
amilz: We could also just use a static keypair. It's not a huge deal, but would just cut CI time and resources a little bit.

catmcgee: not sure i like that. i updated to generate keypair in rust

catmcgee: wdyt?

amilz (approved):
LGTM. 
- Good w/ the keygen/seems fine
- Did not check that the code snippets were migrated over okay
- Spot checked / tested a couple run snippets on the dev server and they worked.
- CI looks like it's [running well](https://github.com/solana-foundation/solana-com/actions/runs/25695942630/job/75444066271?pr=1432#logs)!
- IMO it might be worth pushing a bad commit to one of the commit to check that the CI actually throws an error and then revert before merging.

catmcgee:
> * IMO it might be worth pushing a bad commit to one of the commit to check that the CI actually throws an error and then revert before merging.

did that - failed where it should - reverted it and need your review again :D

amilz (approved):

PR #1447, opened 2026-05-08
Add solfaucet.io to devnet faucets list

MightyVlad:
### Summary of Changes

Adding solfaucet.io to the devnet faucets list.


h4rkl (changes requested):
Requesting changes based on the local PR.md review recommendation.\n\nListing asks users to send mainnet SOL and promises devnet SOL with "no rate limits"; this needs neutral copy and vetting before approval.

PR #1466, opened 2026-05-20
docs: migrate launch docs

ZYJLiu:
migrate content from https://launch.solana.com/docs at https://github.com/solana-foundation/launchonsolana/tree/main/apps/web/content/docs to solana-com and fix/update links


h4rkl (changes requested):
Is this close to being ready bro? Have some big structural changes incoming so clearing PRs.

amilz (commented):
Looks great! Thank you for doing this!

1.  WDYT about calling these `Tools` or `Dev Tools` rather than `Products` (@jacobcreech might have an opinion)
2. On the landing page--would be nice if we could kill the left/right nav buttons at the bottom.

Image

3. I think in an ideal world (maybe we can do this in a separate PR), each tool (e.g., Kora) has its own Sidebar. So if someone is browsing Kora--all the main menus for Kora are top level (and not distracted by say Keychain)--just reduces unnecesary menu nesting imo. Each private sidebar could have an "Other Tools" toggle w/ links to each of the other Tool docs page. 
4. I think our landers for each should be titled {Tool Name} (e.g., Kora) rather than "Introduction" so if you go to `docs/tools/keychain` you see the name in the h1.
5. `docs/products/solana-pay/specification/version1.1` 404'd for me
6. Nit but a few pages e.g. `docs/products/kora/release-notes` don't have any sub-pages but still have a carrot to toggle the menu even though nothing in the menu.

ZYJLiu:
@amilz 
1. Renamed from `Products` to `Tools`
2. Added `hidePageNavigation` frontmatter to hide the left/right nav on `/tools` index.mdx page
3. Updated sidebar to only display pages for the selected tool with an "Other Tools" section to navigate between tools. Is this what you had in mind?

https://github.com/user-attachments/assets/2d3fcd7e-46c2-449b-b5ce-b23fefbd9237

4. Renamed "Introduction" to Tool name on each tool's index.mdx page
5. Renamed mdx file name from `version1.1` to `version1-1`, which was cause the 404
6. Removed dropdown from sections that had no subpages

PR #1467, opened 2026-05-21
cookbook tokens: point Token-2022 fetch examples at mainnet

catmcgee:
### Problem 

 get-token-account, get-token-balance, and get-token-mint hit Token-2022 accounts that the surfnet doesn't lazy-clone cleanly (the rewriter swaps localhost for the hosted surfnet, which returns null for these two)

Other read-only pages still work because their accounts are classic-Token / Token Program itself, which surfnet does have

### Summary of Changes

Switch the URL to api.mainnet.solana.com so the Run button leaves it alone and the code hits mainnet


catmcgee:
closing as we're just going to fix surfpool instead

PR #1471, opened 2026-05-22
feat(ecosystem-data): add Token Creator

massimilianopp:
Adds Token Creator to the ecosystem data registry.

Token Creator is a no-code Solana token launcher for community founders. 
It allows anyone to create an SPL token, add vesting via Streamflow, 
add liquidity via Orca Whirlpool, and share a public token page,
all without writing code.

- Website: https://token-creator.space
- X: https://x.com/TokenCreatorApp


h4rkl (changes requested):
Requesting changes based on the local PR.md review recommendation.\n\nThe ecosystem record uses invalid sector `Tools`; `CompanyProfileSector` only allows values such as `Developer Tools`, so this should fail type checking.

h4rkl:
Need an audit for this to be added. Security getting tighter around these parts. Cheers,

PR #1496, opened 2026-06-02
fix: keep canonical URLs on solana.com

h4rkl:
## Summary
- keep public canonical metadata, hreflang, sitemaps, robots, and structured data on `https://solana.com` across apps
- preserve preview/runtime URL behavior by separating public canonical URLs from deployment `siteUrl` values
- fix proxied docs/i18n redirect host handling so language changes do not bounce to Vercel app roots
- add explicit templates canonical metadata

## Testing
- `pnpm --filter @workspace/i18n test`
- `pnpm --filter solana-com-breakpoint test`
- `pnpm --filter solana-com-media test`
- `pnpm --filter solana-templates check-types`
- `pnpm --filter solana-com-accelerate check-types`
- `pnpm --filter solana-com exec tsc --noEmit`
- Prettier check and `git diff --check`

### DIFF
--- a/apps/accelerate/src/app/metadata.ts
+++ b/apps/accelerate/src/app/metadata.ts
@@ -5,10 +5,10 @@ import faviconSvg from "@solana-com/ui-chrome/assets/favicon.svg";
 import appleTouchIcon from "@solana-com/ui-chrome/assets/apple-touch-icon.png";
 
 export function getBaseMetadata(locale: string = "en"): Metadata {
-  const { siteMetadata, siteUrl, siteIcon, social } = config;
+  const { siteMetadata, publicUrl, siteIcon, social } = config;
 
   return {
-    metadataBase: new URL(siteUrl),
+    metadataBase: new URL(publicUrl),
     title: {
       default: siteMetadata.title,
       template: `%s | ${siteMetadata.title}`,
@@ -21,7 +21,7 @@ export function getBaseMetadata(locale: string = "en"): Metadata {
     openGraph: {
       type: "website",
       locale,
-      url: siteUrl,
+      url: publicUrl,
       siteName: siteMetadata.title,
       title: siteMetadata.title,
       description: siteMetadata.description,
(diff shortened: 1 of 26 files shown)


h4rkl:
Framework audit update:

- next-intl: The existing `localePrefix: "as-needed"` setup is still the right routing mode for default-locale unprefixed URLs. Official docs also confirm middleware/domain matching derives host from `x-forwarded-host` with fallback to `host`, which aligns with preserving the public `solana.com` host behind multi-zone rewrites. `domains` is not a better fit here because the public site uses one canonical domain; the Vercel app hosts are internal zone targets, not locale domains. Docs: https://next-intl.dev/docs/routing/configuration and https://next-intl.dev/docs/routing/middleware
- next-intl upgrade: While checking this, I found the repo catalog was on `next-intl@3.26.5`, affected by the 2026 open redirect advisory for `localePrefix: "as-needed"` and a later advisory patched in `4.9.2`. I bumped to `4.13.0` and confirmed `pnpm audit --prod` no longer reports next-intl advisories. Advisories: https://github.com/advisories/GHSA-8f24-v5vv-gm5j and https://github.com/advisories/GHSA-4c35-wcg5-mm9h
- Next.js: The current multi-zone approach is the documented pattern: route paths between zones with rewrites, use `assetPrefix` for zone assets, and expect hard navigation between zones. The custom `x-forwarded-*` origin handling is still needed because redirects and canonical metadata need the public host, not the internal Vercel target. Docs: https://nextjs.org/docs/pages/guides/multi-zones
- Metadata/canonical: Keeping preview-aware runtime URLs separate from public canonical URLs matches Next metadata guidance: `metadataBase` composes relative canonical/alternate/OpenGraph URLs into fully-qualified URLs. Docs: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase
- Turbo: Turbo only controls task env availability and hashing via `env`/`globalEnv`; it does not participate in runtime redirect origin or canonical URL generation. Docs: https://turborepo.com/docs/crafting-your-repository/using-environment-variables

PR #1499, opened 2026-06-02
docs: clarify versioned message header prefix

### LINKED ISSUE #359: Message header section in docs/core/transactions is confusing regarding versioned transactions
Opening issue related to https://solana.stackexchange.com/q/21089/289

I think the handling of transaction version in the Message Header section here is quite confusing: https://solana.com/docs/core/transactions#message-header

```
The message header uses three bytes to define account privileges:

1 Required signatures and message version(eg. legacy vs v0)
```

To me (and the person who asked that question) this suggests that the first byte somehow encodes both the transaction version and the required signatures. Whereas actually it encodes one of those things, depending if it's a legacy or versioned transaction. It can be used to identify if it's a legacy or versioned transaction, but in the case of a versioned transaction it identifies only the version and the three byte message header will be the following bytes.

This is also the only mention of transaction version on this page. I think it'd make sense to give more information on transaction version on this page (which I wouldn't consider part of the header), and clarify how legacy vs versioned transactions are encoded.

agentmila:
## Summary
- Adds a callout clarifying how the first message byte differs for legacy and versioned transaction messages
- Points readers to the versioned transactions page for the full v0 layout

## Validation

Closes #359

### DIFF
--- a/apps/docs/content/docs/en/core/transactions/transaction-structure.mdx
+++ b/apps/docs/content/docs/en/core/transactions/transaction-structure.mdx
@@ -160,6 +160,15 @@ pub struct MessageHeader {
 
 </WithMentions>
 
+<Callout type="info" title="Legacy and versioned message prefixes">
+  In a legacy transaction message, the first message byte is
+  `num_required_signatures`, followed by the other two `MessageHeader` bytes. In
+  a versioned transaction message, the first byte is a version prefix instead;
+  the three-byte `MessageHeader` starts immediately after that prefix. See
+  [versioned transactions](/docs/core/transactions/versioned-transactions) for
+  the full v0 message layout.
+</Callout>
+
 ![Diagram showing the three parts of the message header](/assets/docs/core/transactions/message_header.png)
 
 ### Account addresses


h4rkl:
Thanks for opening this PR and contributing to solana.com. We appreciate the time you put into it and the community contribution.

I need to close this for now because this repository requires every commit to be signed and verified, and this PR currently includes one or more unsigned commits.

To fix it:

1. Configure commit signing for your GitHub account: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
2. Rework the PR branch so the commit or commits are signed. A common squash flow is:

```bash
# If you do not have an upstream remote yet:
git remote add upstream https://github.com/solana-foundation/solana-com.git

git fetch upstream
git checkout <your-branch>
git rebase -i upstream/main
# Mark extra commits as squash or fixup, then save.
git commit --amend -S --no-edit
git push --force-with-lease
```

If you keep multiple commits, please make sure each one is signed and shows as Verified on GitHub.

After the branch only has signed/verified commits, please reopen this PR, or comment here asking us to reopen it if the button is not available. Opening a fresh PR with signed commits is also fine. Thanks again for contributing.

PR #1510, opened 2026-06-03
feat(news): add category routes

datasalaryman:
### Problem

In preparation for a link to `/news/changelog/`, it would be cool if news article categories had their own routs e.g. 

`/news/defi` for Defi
`/news/developers` for Developers
`/news` is for all news

### Summary of Changes

Made a change so that the news category routes are implemented.

Ensuring the following:
- the scroll reset is set to false
- the hero page renders the most recent article for that category

### DIFF
--- a/apps/media/app/[locale]/news/[...urlSegments]/page.tsx
+++ b/apps/media/app/[locale]/news/[...urlSegments]/page.tsx
@@ -1,4 +1,3 @@
-import React from "react";
 import { notFound } from "next/navigation";
 import Image from "next/image";
 import { reader } from "@/lib/reader";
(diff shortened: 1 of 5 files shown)


h4rkl (changes requested):
The direction is good in principle: making news filters URL-addressable with server-rendered category pages is a sensible improvement. I am requesting changes before approval because the branch is stale relative to main and the URL migration is incomplete.

Required before merge:

- Rebase/update against current main and resolve the merge conflicts. GitHub currently reports this PR as CONFLICTING; a local three-way merge reproduces conflicts in the three news page files.
- Treat /news/:category as the canonical replacement for the existing /news/category/:category URLs. Add permanent redirects for the legacy category URLs, including locale-prefixed variants, preserve the existing changelog exception to /changelog, and update internal navigation, sitemap, structured-data paths, and metadata to the new canonical path. As written, the old route, newsNavHref, the sitemap, and categoryListingMetadata still point at /news/category/..., leaving duplicate URL families and old canonicals.
- Keep category labels separate from category keys/slugs. generateStaticParams currently derives route slugs from category.name, while runtime resolution uses the Keystatic collection key via fetchCategoryByPath; a display-name/key mismatch can produce a statically generated URL that resolves to 404.
- While rebasing, preserve the current news behavior for localized metadata, changelog exclusion/navigation, campaign/masthead rendering, and JSON-LD. Please add coverage for the new routes, legacy redirects, and pagination when navigating between filters.

The existing PR lint, format, and test checks are green (the media Vercel preview is blocked by authorization), so the overall feature direction looks good once these merge and URL-safety issues are addressed.

PR #1574, opened 2026-06-17
docs: Rewrite confidential transfer examples for Token-2022 v11

gitteri:
Now that Token-2022 v11 is live on mainnet, this regrounds the confidential transfer docs in the [confidential-balances-exploration](https://github.com/solana-foundation/confidential-balances-exploration) reference. The old examples were on a two-major-versions-old stack (solana-sdk 2.2 / spl-token-client 0.14) and several used inline proofs, so they no longer match the deployed program.

## Pages updated (the full operation set)

create-mint, create-token-account, deposit, apply-pending-balance, transfer, withdraw. Each now has both a Rust tab and a TypeScript tab.

**Rust:** raw instruction builders on the granular v11 crate stack. Where proofs are involved (configure, transfer, withdraw) they're pre-verified into context state accounts via the ZK ElGamal Proof program and referenced with `ProofLocation::ContextStateAccount`, replacing the old inline `InstructionOffset` proofs. Every Rust example compiles against the v11 reference.

**TypeScript:** the new helpers from `@solana-program/token-2022/confidential` (`getCreateConfidentialTransferAccountInstructionPlan`, `getConfidentialTransferInstructionPlan`, `getConfidentialWithdrawInstructionPlan`, `getApplyConfidentialPendingBalanceInstructionFromToken`) with recoverable ElGamal/AES key derivation, plus `getCreateMintInstructionPlan` and `getConfidentialDepositInstruction` for the proof-less operations. Every TS example typechecks against the published packages.

Also updated the proof-flow prose/diagrams, removed a stale duplicate `[dependencies]` block on create-mint, switched create-mint to configure an auditor (rather than `None`), and replaced the run instructions: confidential transfers need the ZK ElGamal Proof program, which is enabled on mainnet, devnet, and a mainnet-forking local validator like [Surfpool](https://surfpool.run), but not the stock `solana-test-validator`.

## Notes

- The Rust `Cargo.toml` pins the `solana-client` `4.0.0-rc.0` line plus granular crates rather than the `solana-sdk` umbrella: stable `solana-client 4.0.0` caps `solana-instruction` below what token-2022 v11 (`solana-system-interface 3.2`) needs. Each manifest has a comment noting it collapses back to `solana-sdk` once a compatible stable client ships.
- TS examples are focused on the confidential-transfer logic; the `@solana/kit` client/instruction-plan setup is left to the reader's standard kit setup.
- English source only; translations regenerate.

Follow-ups left out: the `index.mdx` overview and `issuer-guide.mdx` (and the new `integration-guide.mdx` that just landed on main), plus a few proof-type doc links that still point at the renamed `zk-token-sdk` crate.


gitteri:
@greptile review

The decrypt_u32 range-limit findings on `withdraw-tokens.mdx` and `apply-pending-balance.mdx` are fixed (both now read the available balance from the AES-encrypted `decryptable_available_balance`).

On the remaining transfer note: `transfer-tokens.mdx` does not contain an `apply_pending_balance_for_account` helper or any `decrypt_u32` call. The transfer example reads the available balance via `current_decryptable.decrypt(&sender_aes)` (AES) and passes the ElGamal ciphertext to `transfer_split_proof_data` only for proof generation, so there is no decrypt_u32 range-limit issue in that file.

catmcgee (approved):

PR #1581, opened 2026-06-18
fix: i18n source locale overwrites

h4rkl:
## Summary
- restore authored English docs content from #1573/#1524 that was regressed by the Lingo #1572 update
- restore the confidential transfer integration guide card/nav entry and remove stale availability warnings
- prevent future Lingo runs from treating `en` as a generated target locale
- add a source-locale verification guard to local i18n scripts and the GitHub i18n workflow

## Validation
- `pnpm i18n:verify-source-locales`
- `pnpm --filter @workspace/i18n lint`
- `pnpm --filter solana-docs postinstall`
- `git diff --check`

### DIFF
--- a/apps/docs/content/docs/en/tokens/extensions/confidential-transfer/meta.json
+++ b/apps/docs/content/docs/en/tokens/extensions/confidential-transfer/meta.json
@@ -7,6 +7,7 @@
     "apply-pending-balance",
     "withdraw-tokens",
     "transfer-tokens",
+    "integration-guide",
     "issuer-guide"
   ]
 }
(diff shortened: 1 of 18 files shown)


apps/docs/content/docs/en/tokens/extensions/confidential-transfer/meta.json
```diff
@@ -7,6 +7,7 @@
     "apply-pending-balance",
     "withdraw-tokens",
     "transfer-tokens",
+    "integration-guide",
```
gitteri, line 10: i don't actually see the integration guide in the change list

h4rkl: It was never overwritten - exists here in branch https://github.com/solana-foundation/solana-com/blob/6c76353bf176b3316b1af555173cf1651c58810a/apps/docs/content/docs/en/tokens/extensions/confidential-transfer/integration-guide.mdx

gitteri (approved):

PR #1691, opened 2026-07-14
docs(guides): clarify token_acl auto-thaw is SDK-side (not Phantom)

### LINKED ISSUE #1686: Docs inaccuracy: acl.mdx names Phantom as auto-thawing token_acl mints, but Phantom doesn't support it (sRFC-37 Token ACL)
## Summary

The Token ACL (sRFC-37) guide states that wallets like Phantom automatically detect the `token_acl` metadata field and append a permissionless thaw instruction when creating token accounts. Based on direct confirmation from Phantom's team, this is **not currently the case** for Phantom, so the guide may mislead developers relying on this behavior.

## Location

`apps/docs/content/guides/advanced/acl.mdx` — section **"Token Metadata for Auto-Detection"**.

The guide currently states (paraphrased): adding the `token_acl` metadata field is crucial for wallet integration, and when wallets like Phantom (or SDKs like `@solana/token-helpers`) see this field, they automatically include thaw instructions when creating token accounts.

Rendered page: https://solana.com/developers/guides/advanced/acl#token-metadata-for-auto-detection

Image

## What I observed

I created a Token-2022 mint with `DefaultAccountState = Frozen`, a Token ACL config, and an ABL Gate Program allow list (`ListConfig`, mode `Allow`) with the recipient wallet added. The mint's metadata includes the `token_acl` field pointing to the Gate Program.

When sending this token from Phantom (Extension v26.20.0) to an allow-listed wallet that has no ATA yet, Phantom created the ATA but did **not** append a `ThawPermissionless` instruction. The ATA stayed frozen and the transfer failed with **Error `0x11` (`AccountFrozen`)** from Token-2022 — consistent with the ATA never being thawed.

## Confirmation from Phantom

I raised this with Phantom's developer support. Their team (Phantom DevRel) confirmed that Phantom **does not currently support this flow**, that the observed behavior is expected, and that it is **not a misconfiguration** on my side. They suggested submitting a feature request to get it on their roadmap.

## Questions / suggested fix

1. Since Phantom is named explicitly in the guide but does not currently support auto-detection + permissionless thaw, could this section be updated to reflect the current state?
2. **Which wallets, if any, actually support this today?** It would be very helpful for the guide to list wallets with verified support, rather than naming Phantom as an example if support isn't live yet.
3. If no wallet currently ships this automatically, could the guide clarify that automatic `token_acl` detection is **aspirational / SDK-side only** (e.g. via `@solana/token-helpers` when the integrating app builds the transaction), and that wallets do not append the thaw instruction on their own yet?

agentmila:
## Summary
- Corrects the Token ACL guide so it no longer states that wallets like **Phantom** automatically detect `token_acl` and append a permissionless thaw when creating token accounts.
- Clarifies that auto-detection is primarily an **SDK / integrating-app** behavior (e.g. `@solana/token-helpers`).
- Notes that **Phantom does not currently support** this wallet-side flow (reporter confirmed with Phantom DevRel).

Fixes #1686

## Why
The previous wording caused real integration failures: Phantom created a frozen ATA without thaw, and transfers failed with Token-2022 `AccountFrozen` (`0x11`).

## Changes
- `apps/docs/content/guides/advanced/acl.mdx`
  - Key concept + callouts updated
  - CLI workshop comments no longer say "wallet auto-detection" as a guarantee

## Validation
- Diff reviewed against issue report and rendered section path `guides/advanced/acl.mdx`
- No code/runtime changes (docs-only MDX)

## Notes
- No verified public list of wallets that currently ship this auto-thaw was available; docs intentionally avoid naming unsupported wallets as examples.

### DIFF
--- a/apps/docs/content/guides/advanced/acl.mdx
+++ b/apps/docs/content/guides/advanced/acl.mdx
@@ -230,8 +230,8 @@ Token ACL consists of three main components:
    issuer intervention, as long as the Gate Program approves.
 
 4. **TokenMetadata Integration**: Adding a `token_acl` field to your mint's
-   metadata enables automatic detection by wallets and SDKs like
-   `@solana/token-helpers`.
+   metadata enables automatic detection by SDKs like `@solana/token-helpers`
+   when *they* build create-account transactions.
 
 <Callout type="info" title="Auto-Detection with TokenMetadata">
 
@@ -239,6 +239,15 @@ When you add a `token_acl` field to your mint's TokenMetadata extension pointing
 to the Gate Program address, SDKs like `@solana/token-helpers` can automatically
 detect Token ACL mints and include thaw instructions when creating token
 accounts.
+
+This is primarily an **SDK / integrating-app** behavior today. Do **not** assume
+that wallets automatically append a permissionless thaw when a user receives
+tokens or creates an ATA. In particular, **Phantom does not currently support**
+auto-detecting `token_acl` and adding thaw instructions on account creation
+(confirmed with Phantom DevRel as of the report in
+[#1686](https://github.com/solana-foundation/solana-com/issues/1686)). If your
+flow depends on a thawed ATA, include the thaw instruction yourself (or use an
+SDK that does).
 
 </Callout>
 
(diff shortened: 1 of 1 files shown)


h4rkl:
Thanks for the careful report and fix. A current, targeted PR now covers this issue, so we are closing this superseded version to keep review focused. We appreciate your contribution.

PR #1773, opened 2026-07-23
docs: Add transfer hook integration guide for wallets

gitteri:
## Summary

Adds a wallet/dapp/custodian integration guide for the Transfer Hook extension under `tokens/extensions`, mirroring the confidential-transfer extension's issuer-vs-integration split. Covers detecting the extension on a mint, resolving the `ExtraAccountMetaList` PDA (both via the high-level `createTransferCheckedWithTransferHookInstruction` helper and the lower-level resolver functions), why simulating before sending matters (hook rejections and stale extra-account state both fail onchain, not gracefully), compute/setup implications, and the read-only-accounts-during-CPI safety design. Issuer-facing guidance (writing/deploying a hook program) is explicitly out of scope, per the ticket.

Registered in `apps/docs/content/docs/en/tokens/extensions/meta.json` right before `confidential-transfer`.

This is a first-pass draft per solutions-bot's task-execution workflow, not a final/polished version.

## Notes for reviewers

- Mirrored shape/rigor from `tokens/extensions/confidential-transfer/integration-guide.mdx` (closest existing precedent for an extension's integration-only guide)
- Used the rich frontmatter format per the going-forward standard, even though the neighboring confidential-transfer page still uses the minimal form
- Verified `getTransferHook`, `getExtraAccountMetaAddress`, `resolveExtraAccountMeta`, `getExtraAccountMetas`, `addExtraAccountMetasForExecute`, and `createTransferCheckedWithTransferHookInstruction` against the actual `@solana/spl-token` / `token-2022` client source rather than relying on memory
- Flagged inline that the Kit-based `@solana-program/token-2022` client has low-level instruction builders but no equivalent high-level extra-account resolver yet
- Ran `pnpm --filter solana-docs lint` (frontmatter guard, eslint, prettier) clean, and confirmed the page renders (200) and the sidebar entry appears via a local dev server before opening this PR

Refs: ECO-413


amilz (dismissed):

apps/docs/content/docs/en/tokens/extensions/transfer-hook-integration.mdx
```diff
@@ -0,0 +1,510 @@
...
+prerequisites:
+  - /docs/tokens/extensions
+related:
+  - /docs/tokens/extensions/transfer-hook
+  - /docs/tokens/extensions/permanent-delegate
+  - /docs/tokens/extensions/transfer-fees
+  - /docs/tokens/extensions/confidential-transfer/integration-guide
+---
+
+# Supporting Transfer Hooks on Solana
```
amilz: this kind of leaves. you with 2 titles
Image

apps/docs/content/docs/en/tokens/extensions/transfer-hook-integration.mdx
```diff
@@ -0,0 +1,510 @@
...
+        simulation.value.logs?.join("\n")
+    );
+  }
+
+  // 4. Sign and send only after a successful simulation.
+  return sendAndConfirmTransaction(connection, transaction, [payer, owner]);
+}
+```
+
+#### Detecting the extension
```
amilz: These are all nested under web3js but hit webjs & kit 

Image

apps/docs/content/docs/en/tokens/extensions/transfer-hook-integration.mdx
```diff
@@ -0,0 +1,510 @@
...
+  if (simulation.value.err) {
+    throw new Error(
+      `Transfer simulation failed: ${JSON.stringify(simulation.value.err)}\n` +
+        simulation.value.logs?.join("\n")
+    );
+  }
+
+  // 3. Sign only after a successful simulation, then send.
+  const signedMessage = await signTransactionMessageWithSigners(message);
+  await sendAndConfirmTransactionFactory({ rpc, rpcSubscriptions })(
```
amilz, line 215: i needed this

```suggestion
  assertIsTransactionWithBlockhashLifetime(signedTransaction)
  await sendAndConfirmTransactionFactory({ rpc, rpcSubscriptions })(
```

apps/docs/content/docs/en/tokens/extensions/transfer-hook-integration.mdx
```diff
@@ -0,0 +1,510 @@
...
+Every hook-enabled transfer needs the standard four transfer accounts (source,
+mint, destination, owner/authority) plus whatever the `ExtraAccountMetaList`
+account for that mint specifies. The list is a PDA derived from the hook
+program:
+
+```ts title="derive-extra-account-meta-list.ts"
+// Kit (@solana-program/token-2022)
+import { findExtraAccountMetaListPda } from "@solana-program/token-2022";
+
+const [extraAccountMetaListPda] = await findExtraAccountMetaListPda(
```
amilz, line 369: Kit gets kind of grumpy unless it knows about the extension. I think it's probably fine to exclude this but just flagging in case you think it's important to include.

```ts
const mintAccount = await fetchMint(rpc, mint);
const transferHook = unwrapOption(mintAccount.data.extensions)?.find(
  (extension) => extension.__kind === "TransferHook"
);

if (transferHook) { 
// ...
```

amilz (dismissed):
Looks great!

apps/docs/content/docs/en/tokens/extensions/transfer-hook-integration.mdx
gitteri: Fixed in 1b3b53c31: dropped the redundant `# Supporting Transfer Hooks on Solana` H1 so the frontmatter `title` is the only page title (matches transfer-fees / permanent-delegate / transfer-hook, which all start at `##`).

apps/docs/content/docs/en/tokens/extensions/transfer-hook-integration.mdx
gitteri: Fixed in 1b3b53c31: promoted Detecting the extension / Resolving the extra accounts / Simulating before sending to top-level `##` sections so they no longer nest under Web3.js in the TOC. They cover both clients, so they now sit as siblings of the Kit and Web3.js drop-in functions rather than under one of them.

apps/docs/content/docs/en/tokens/extensions/transfer-hook-integration.mdx
gitteri: Leaving this out for now per your call: the Kit drop-in path uses `getTransferCheckedWithTransferHookInstructionAsync`, which fetches the mint and detects the hook internally, so the guide does not show a manual `fetchMint` + `unwrapOption` detect step on the Kit side. Happy to add the explicit Kit detection snippet if you think it earns its place.

amilz (approved):

PR #1793, opened 2026-07-28
Add Vultisig to the wallet directory

realpaaao:
This PR adds [Vultisig](https://vultisig.com) to the wallet directory (`packages/ecosystem-data/src/wallets/wallet-data.ts`).

Vultisig is a seedless, open-source, self-custodial MPC wallet. Vault shares are split across the user's own devices (2-of-2 minimum, larger thresholds supported): no seed phrase, no single point of failure. SOL and SPL tokens are supported for send/receive/swap plus SOL staking, on iOS, Android, browser extension, and desktop (Windows/macOS/Linux). Source: https://github.com/vultisig

The change follows the existing patterns: entry in alphabetical position with taxonomy-valid category/platform/feature keys, a new `vultisigSubmission` verification batch in `metadata.ts` (as the file's comment prescribes for new verification dates), and a 128×128 webp icon matching the existing icon format.

## receipts

- `tsc --noEmit` on `packages/ecosystem-data` passes with the new entry — validates the taxonomy keys, the `lastVerified` batch wiring, and the icon import.


h4rkl:
Thanks for the contribution. The live wallet directory already contains an independently verified Vultisig record, so this PR is no longer needed. We appreciate the submission.

PR #1806, opened 2026-07-30
Fix broken links and redirects

catmcgee:
- Repair moved and retired internal routes with locale-aware redirects, including Markdown variants.
- Update broken docs and developer-page links to their current authoritative destinations.


catmcgee:
Reopening this from an org branch so the Vercel previews can run.

PR #1847, opened 2026-08-05
docs: resolve documentation audit findings

catmcgee:
## Summary

Audit: https://falconer.com/audits/solana.html

- standardize English docs and samples on the canonical api.mainnet.solana.com RPC host
- reconcile simulation-log nullability, modernize the exchange guide RPC flow, repair stale links, and align inconsistent examples
- align Kora commands and configuration with the stable v2.0.5 implementation
- keep Local Installation detail guides in the Resources sidebar and navigation context while preserving their existing URLs

## Audit review notes

A few reported items were intentionally not changed after source verification:

- get_payment_instruction is a client helper, not a Kora server method that belongs in enabled_methods
- web3.js does not expose getBlocksWithLimit, so that page should not add a web3.js tab
- getAccountInfo still has a binary protocol default; the docs now distinguish that compatibility default from the recommended explicit base64 encoding


apps/docs/content/docs/en/defi/exchange.mdx
```diff
@@ -963,12 +967,14 @@ You can then check the installed version to verify
 spl-token --version
 ```
 
-Which should result in something like
+This prints the installed version, for example:
 
 ```text
-spl-token-cli 2.0.1
+spl-token-cli 3.0.0
```
amilz, line 973: if changing this, might as well put latest, 5.6.1? is v3 sufficient for confidential tokens?
https://crates.io/crates/spl-token-cli

packages/ui-chrome/src/developers-nav.tsx
```diff
@@ -64,6 +66,7 @@ export function DevelopersNav({
                   "/docs/tokenization",
                   "/docs/defi",
                   "/docs/tools",
+                  localInstallationResourcesPath,
```
amilz, line 69: this is odd but fine i guess

amilz (dismissed):
Thanks for doing this. Couple of nits.

apps/docs/content/docs/en/defi/exchange.mdx
catmcgee: updated to latest

amilz (dismissed):

amilz (approved):

PR #1859, opened 2026-08-06
docs(surfpool): document the kit plugin

amilz:
## Summary

- New SDK page for the `@solana/surfpool/kit` entry point — embedded and attach modes, what the plugin installs, lifecycle, the typed cheatcodes RPC, and composing with Kit program plugins. The Surfpool docs previously had no Kit coverage at all.
- Includes a `setAccount` example that encodes an SPL mint with `getMintEncoder()` and hex-encodes it, standing up an initialized mint with supply and no transactions.
- Links the page from the tool landing page, SDK overview, installation, and JS reference.
- Records the Node 20.18+ floor the Kit packages declare (`@solana/surfpool` itself is 18+, `@solana-program/token` is 24+).

## Test Plan

- `pnpm lint --filter solana-docs` passes (ESLint, prettier, frontmatter guard).
- Every code snippet on the page was typechecked as one file under `tsc --strict` against the installed `@solana/kit` v7, `kit-plugin-rpc`, `kit-plugin-signer`, `@solana-program/token` and `@solana-program/system`.
- The codec/`setAccount` example was executed against a real embedded Surfnet and reads back through `fetchMint` (decimals 6, supply 1_000_000_000n, owner = Token program).

https://solana-com-docs-git-docs-too-631-surfp-152173-solana-foundation.vercel.app/docs/tools/surfpool/sdk/kit-plugin

Closes TOO-631

### DIFF
--- a/apps/docs/content/docs/en/tools/surfpool/sdk/kit-plugin.mdx
+++ b/apps/docs/content/docs/en/tools/surfpool/sdk/kit-plugin.mdx
@@ -1,0 +1,375 @@
+---
+title: Kit Plugin
+description:
+  Run an embedded Surfnet behind a Solana Kit client. One .use(surfpool()) gives
+  you a pre-funded payer, the full RPC stack, and typed cheatcodes.
+---
+
+`@solana/surfpool/kit` runs a Surfnet — a local, Solana-compatible network —
+inside your test process and hands back a
+[Solana Kit](https://www.solanakit.com/) client already pointed at it. One
+`.use(surfpool())` replaces the RPC plugin you would normally reach for
+(`solanaLocalRpc()`, `litesvm()`) and adds a pre-funded payer plus Surfpool's
+cheatcodes:
+
+```ts
+import { createClient } from "@solana/kit";
+import { surfpool } from "@solana/surfpool/kit";
+
+const client = await createClient().use(surfpool());
+
+const slot = await client.rpc.getSlot().send();
+await client.cheatcodes.timeTravel({ absoluteSlot: 1_000_000n }).send();
+```
+
+No port to pick, no payer to generate and fund, and no separate `surfpool start`
+process to manage. New to the SDK? Start with the
+[Overview](/docs/tools/surfpool/sdk/overview).
+
+## Which Entry Point You Want
+
+| Entry point                       | Reach for it when                                                                                         |
+| --------------------------------- | --------------------------------------------------------------------------------------------------------- |
+| `surfpool()`                      | **Default for tests.** An isolated Surfnet per test file, with a Kit client already wired up.             |
+| `surfpool({ rpcUrl })`            | A long-lived `surfpool start` instance is shared across processes, or your platform has no native binary. |
+| `surfnetCheatcodes()`             | You already have a client and only want cheatcodes on it.                                                 |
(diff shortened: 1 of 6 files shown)


apps/docs/content/docs/en/tools/surfpool/sdk/kit-plugin.mdx
```diff
@@ -0,0 +1,375 @@
...
+- **A [supported platform](/docs/tools/surfpool/sdk/installation)** (macOS,
+  Linux x86-64) for embedded mode, which loads a native binary. Elsewhere, use
+  attach mode.
+- **Familiarity with Kit's plugin composition** — clients are built by chaining
+  `.use()` calls, and each plugin adds properties to the client.
+
+## Installation
+
+```bash
+npm install --save-dev @solana/kit @solana/kit-plugin-rpc @solana/kit-plugin-signer
```
MicaiahReid: Since this page is for the surfpool kit plugin, should the install script include `@solana/surfpool`?
```suggestion
npm install --save-dev @solana/kit @solana/kit-plugin-rpc @solana/kit-plugin-signer @solana/surfpool
```

apps/docs/content/docs/en/tools/surfpool/sdk/kit-plugin.mdx
```diff
@@ -0,0 +1,375 @@
...
+# or
+pnpm add -D @solana/kit @solana/kit-plugin-rpc @solana/kit-plugin-signer
```
MicaiahReid:
```suggestion
pnpm add -D @solana/kit @solana/kit-plugin-rpc @solana/kit-plugin-signer @solana/surfpool
```
Same as above

apps/docs/content/docs/en/tools/surfpool/sdk/kit-plugin.mdx
```diff
@@ -0,0 +1,375 @@
...
+  .use(payer(myPayer))
+  .use(surfpool({ rpcUrl: "http://127.0.0.1:8899" }));
+```
+
+Three differences from embedded mode:
+
+- **The client must already have a `payer`.** Attach mode has no access to the
+  running instance's payer secret key, so it installs none. Fund whichever
+  signer you supply with `client.cheatcodes.setAccount(...)` or the running
+  instance's own faucet.
```
MicaiahReid, line 350: Not a docs change, a potential future change to the kit client, but could we allow:
```js
const client = createClient()
  .use(payer(myPayer))
  .use(surfpool({ 
       rpcUrl: "http://127.0.0.1:8899",
       airdropAddresses: [myPayer, ...]
    })
  );
```
which, upon connection to the existing surfnet, sends an airdrop request to the network automatically?
This would prevent the user from having to make a separate cheatcode call

MicaiahReid (changes requested):
This is really great! Thanks @amilz!

Something interesting for the future, which would probably be a separate page, but it would be helpful to see "when do I use surfpool vs litesvm vs solanaLocalRpc for my tests" be answered. There are advantages/disadvantages to each - it would be great to see the cases where surfpool particularly shines in tests

apps/docs/content/docs/en/tools/surfpool/sdk/kit-plugin.mdx
amilz: added a ticket. will get on this

MicaiahReid (approved):
Awesome stuff!

PR #1933, opened 2026-08-19
Serve RPC latency data without unstable_cache to stop wedged snapshots

niks3089:
Removes the `unstable_cache` layer from `/api/rpc/data` and `/api/rpc/filters`.

**Why:** `unstable_cache` entries with `revalidate: 60` silently stop refreshing and serve hours-old snapshots (a `getBlock_recent`/pit key was observed frozen for 6.5h while all six underlying Prometheus queries succeeded in under a second). Partner-visible fallout: providers appearing/disappearing when switching timeframes, and providers missing from Avg Latency — the Avg Latency instant query is timeframe-independent but was cached under a timeframe-keyed entry, so each timeframe served a snapshot from a different moment.

**What still shields Grafana:** the CDN cache (`s-maxage=60` / `s-maxage=300` + stale-while-revalidate) and the existing per-instance in-flight request dedup, now keyed directly by the query cache key. Grafana Cloud handles the residual origin misses fine.

Not touched: `/api/rpc/sender/data` uses the same pattern, but its ClickHouse backend has availability issues where the data cache doubles as a fallback — can get the same treatment separately if wanted.

Note: providers with a genuinely empty trailing-5m window still won't get an Avg Latency bar; that's now honest realtime behavior rather than stale-snapshot roulette. Rendering an explicit per-provider "no data" state would be a frontend change out of scope here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


h4rkl:
Superceded by [2031](https://github.com/solana-foundation/solana-com/pull/2031)

PR #1938, opened 2026-08-20
fix(i18n): harden Lingo docs recovery

h4rkl:
## Problem

The localization job could get stuck or finish with missing docs translations. Some generated docs frontmatter also came back with backticks that made validation fail.

## Solution

Make each localization run cancel older runs on the same branch, add a long but fixed timeout, and teach the Lingo runner to retry only the source files that failed. After Lingo runs, normalize translated docs frontmatter by removing backticks and refreshing matching lockfile hashes.

## Testing

TODO: Confirm CI localization workflow passes.

### Summary of Changes

- Updated i18n workflow concurrency and timeout settings.
- Added docs frontmatter normalization after Lingo generation.
- Reworked `scripts/i18n/run.mjs` to support async Lingo execution, wait timeouts, targeted retries, recovery pulls, and backfill mode.
- Added `scripts/i18n/sanitize-docs-frontmatter.mjs` for translated docs MDX frontmatter cleanup.

---

### Change classification (SDLC §2)

- [ ] **Critical** — auth, secrets/key handling, fund movement, deploy/CI
      security gates, or anything that changes who can do what
- [x] **Standard** — production-facing, non-critical (features, API/route
      changes, dependency updates, monitoring/config)
- [ ] **Low** — no security impact (docs, tests, dev tooling, content)

### Security checklist

- [ ] No secrets, tokens, or credentials in source, env files, or CI logs
      (use Doppler / `NEXT_PUBLIC_*` only for values safe to ship to the
      browser).
- [ ] Input from users or external services is validated before use; no
      `dangerouslySetInnerHTML` / unsanitized HTML on untrusted input.
- [ ] New or updated dependencies are justified below, and `pnpm audit`
      passes (no new High/Critical advisories).
- [ ] Errors are handled explicitly — no silent failures on production paths.
- [ ] For **Critical** changes: a trust-boundary / threat-model note is
      included below, and a second reviewer has been requested.


h4rkl:
re @Greptile-app

1 — "scoped commands can still alter unrelated translation buckets": true, but intentional and load-bearing. The code says why at scripts/i18n/run.mjs:369: current Lingo CLI releases treat positional file patterns as force/new-file scopes, so passing a scope to push would force-retranslate files instead of doing an incremental sync. Incremental pushes therefore must be config-wide, with the lockfile limiting the work to changed sources. Given that, the repo-wide coverage guard and backfill (run.mjs:382-399) are consistent rather than a leak: the push's blast radius is already the whole config, so any gap the guard surfaces was inside this run anyway. Greptile's implied fix — restricting coverage/backfill to the requested scope — would actually be worse: it would let a run exit green while leaving known-missing targets in other buckets, and verify-target-coverage.mjs has no scope parameter to do it with. The backfill only creates genuinely missing targets, so the "alteration" of unrelated buckets is filling holes, not rewriting content.

2 — "recovery accepts MDX on existence + changed hash rather than completeness": stale. That was the state before the head commit (c61e282c6 fix(i18n): require completed resume for push recovery). As of run.mjs:256-298, recovery clears a failed push only when all three hold: lingo resume exits 0 (Lingo itself confirms the interrupted run finished — this is the completeness proof, and the comment at run.mjs:270 states exactly the objection Greptile raises), the pull succeeds, and every target the CLI reported as failed was rewritten to different bytes. The hash check is a supplementary guard on top of the resume confirmation, not the acceptance criterion. If resume can't confirm completion, the failed status is kept even when the pull salvaged files (run.mjs:273-278). Downstream, every push path — recovered or not — still passes through the repo-wide coverage check and the docs frontmatter sanitize/verify pair (run.mjs:385-403).

h4rkl:
Fixing:
- @vercel: comment — [vc]: #MvppVTY74hVDrr4bbmZ0hM4iWN/oW96IdfEI90a2AkI=:eyJpc01vbm9yZXBvIjp…
- @greptile-apps: comment — <h3>Greptile Summary</h3>
- @h4rkl: comment — re @Greptile-app

Will be in the next commit.

h4rkl:
Fixing:
- @vercel: comment — [vc]: #MvppVTY74hVDrr4bbmZ0hM4iWN/oW96IdfEI90a2AkI=:eyJpc01vbm9yZXBvIjp…
- @greptile-apps: comment — <h3>Greptile Summary</h3>
- @h4rkl: comment — re @Greptile-app
- @greptile-apps: comment — Sounds good, thanks for the heads up! Let me know when the commit is up…

Will be in the next commit.

amilz (approved):

PR #1910, opened 2026-08-14
docs(payments): reorganize payment and agentic payment guides

catmcgee:
## Summary

- add protocol-neutral agentic payment guidance with MPP coverage for Solana charges, sessions, verification, and production safeguards
- add an x402 V2 overview and payment walkthrough covering roles, headers, network identifiers, SDK examples, and spend policies
- move Become an x402 Facilitator into Resources across every docs locale
- replace the stale Kora x402 V1 guide with current V2 architecture and deployment guidance
- consolidate duplicate payment-button, Pay.sh, and subscription guides into canonical pages
- preserve retired payment and x402 URLs with permanent redirects

Fixes DEV-966

PR #2002, opened 2026-08-27
docs: deduplicate payment integration guides

catmcgee:
## Summary

- make the commerce kit paymentbutton reference the single canonical component page
- combine the pay sh setup and subscription material into one current guide
- remove duplicate sidebar entries and update affected payment links
- preserve the three retired urls with permanent redirects


catmcgee:
consolidated into #1910

PR #2007, opened 2026-08-28
Add installation scripts CI to ensure that the commands are up to date

datasalaryman:
### Problem

we had an issue awhile back where the recommended way to install AVM broke and had to be replaced. we received no notification or run indicating that the script needed to be updated. 

### Summary of Changes

Extract all terminal commands using a python script. Run via CI.


h4rkl:

The current head has 3 commit(s), and GitHub reports no signature failure for the commit verification check.

PR #2020, opened 2026-08-31
fix(web): reflect the selected timeframe in RPC latency chart captions

McSim85:
## Problem

On `/data?tab=rpc`, the P50/P95/P99 latency chart captions are static strings with the default timeframe baked in: change the Timeframe selector to 1h (or 24h, 7d, …) and the charts show the new range while their captions still read "Hourly P50 · last 6 hours · lower is better". Reproduced on Brave and Firefox.

The captions come from `dataDashboard.charts.rpc-p*-latency.caption` in `packages/i18n/messages/web/en/common.json` and are rendered by `getChartCaption()` with no knowledge of the selected timeframe.

## Changes

- The three captions take a `{timeframe}` ICU parameter: "Hourly P50 · last {timeframe} · lower is better".
- `rpcTimeframe` is threaded from `SolanaDataDashboard` through `ChartGrid` into `ChartCard`, and `getChartCaption()` accepts optional interpolation values. The label shown is the selector's own label (30m/1h/6h/24h/7d/30d/90d/1y), so caption and selector always agree.
- "Hourly" is kept as-is deliberately: the latency quantiles are computed over a fixed 1-hour window (`RPC_LATENCY_QUANTILE_RANGE` in `apps/web/src/lib/rpc/server.ts`) at every timeframe, so that half of the caption is accurate.

Captions without placeholders (all other charts) are unaffected — unused interpolation values are ignored.

## Testing

- `pnpm --filter solana-com check-types` clean
- `pnpm --filter solana-com` vitest suite: 33 files, 242 passed / 18 skipped
- `pnpm --filter @workspace/i18n test`: 21 passed


McSim85:
Good catch — the 18 translated locales carried their own literal "last 6 hours" and would have kept overriding the parameterized English template. Fixed in c24fe03: every locale's three RPC captions now take the same `{timeframe}` placeholder, with the substitution placed to fit each language's word order (e.g. `過去{timeframe}`, `{timeframe} terakhir`, `dernières {timeframe}`). 54 strings changed, verified with the @workspace/i18n test suite.

McSim85:
Note on the locale commit vs. the Lingo pipeline: we're aware the i18n workflow re-translates changed en keys on merge (checksum-driven via .lingo/lock.json) and will normalize these three keys in its next "i18n: Lingo Translation Update" PR. The hand-parameterized translations are included anyway so the fix is complete in every locale from the moment this merges, with no window where non-English visitors still see the literal "last 6 hours". Feel free to drop c24fe03 if you'd rather let Lingo produce the target strings.

h4rkl:

The current head has 2 commit(s), and GitHub reports `unsigned` for the commit verification check.

McSim85:
Done — squashed to a single signed commit and force-pushed; content is identical to the two commits you reviewed (caption parameterization + all 18 locales). Base is unchanged; it merges cleanly onto current `main`, including the Lingo update in #1994.

McSim85:
The red Tests job is unrelated to this change, it seems like: `@workspace/i18n` › "falls back to English for new Breakpoint route messages" asserts that `es` has no translation for `breakpoint.travel.visas.checkRequirements`, and the Lingo update in #1994 added one, so the test fails on any PR run against current `main`. #2027 rewrites that test to not depend on the live locale files. Everything else here is green (format, all lint jobs, Greptile, the web suite inside the same Tests job).

PR #2027, opened 2026-09-02
fix: repair flaky i18n test and patch 2 high-severity CVEs

BlocksOnAChain:
## Summary

Two CI failures I found while verifying an unrelated PR, unrelated to any specific feature work:

- **`@workspace/i18n`'s flaky test**: "falls back to English for new Breakpoint route messages" asserted hardcoded English strings against the live Breakpoint content catalog for two keys that, when the test was written, hadn't been translated into Spanish yet. Both have since been fully translated (Spanish now has 100% parity with English for Breakpoint), so the test's fixture assumption went stale and started failing correctly, since the real message is now the Spanish string rather than the English fallback. Rewrote it to use synthetic data via `deepMergeMessages` directly, matching the pattern the two sibling tests in this same file already use, so it verifies the same fallback mechanism without depending on the live catalog's translation completeness at any given time (which will only keep changing as translators catch up).
- **`pnpm audit --audit-level high --prod`** (the exact command CI runs) was failing on:
  - `GHSA-c83g-rgw3-j3cx` / `GHSA-73wf-gq98-2v4g` (`browserslist`, transitive via webpack in every app)
  - `GHSA-j4fx-xxwh-2485` (`jsondiffpatch`, a direct dependency of the `ai` package `apps/docs` uses)

  Added `pnpm.overrides` entries pinning both to their patched releases, matching the existing pattern in this file for `esbuild`/`postcss`/`nanoid`/etc. Verified the resulting lockfile diff touches only these two packages' own dependency subtrees, no unrelated version bumps.

## Test plan

- [x] `pnpm audit --audit-level high --prod` exits 0 (was exit 1)
- [x] `pnpm exec turbo run test --filter='!@workspace/docs-examples'` is 9/9 green (was failing on `@workspace/i18n#test`)
- [x] Full monorepo `check-types`: 15/15 clean
- [x] Full monorepo `lint`: 13/13 clean (pre-existing warnings only, 0 errors)
- [x] Confirmed the lockfile diff is scoped to `browserslist`/`jsondiffpatch`'s own dependency trees only

### DIFF
--- a/packages/i18n/src/messages.test.ts
+++ b/packages/i18n/src/messages.test.ts
@@ -118,19 +118,24 @@ describe("@workspace/i18n messages", () => {
     expect(messages).toHaveProperty("cookie-consent");
   });
 
-  it("falls back to English for new Breakpoint route messages", async () => {
-    const messages = await loadMergedMessages({
-      app: "breakpoint",
-      locale: "es",
-    });
+  it("falls back to English for new Breakpoint route messages", () => {
+    // Uses synthetic data rather than the live breakpoint catalog: a key
+    // that's still untranslated today has no guarantee of staying that way
+    // as translators catch up, which would make this test flaky over time.
+    const messages = deepMergeMessages(
+      {
+        breakpoint: {
+          travel: {
+            visas: { checkRequirements: "Check visa or ETA requirements" },
+          },
+        },
+      },
+      { breakpoint: { travel: { visas: {} } } },
+    );
 
     expect(messages).toHaveProperty(
       "breakpoint.travel.visas.checkRequirements",
       "Check visa or ETA requirements",
-    );
-    expect(messages).toHaveProperty(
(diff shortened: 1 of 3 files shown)


BlocksOnAChain:
Closing as superseded — both fixes this branch targeted have since landed on `main` independently, via different (and in the case of the audit fix, more narrowly-scoped) approaches:

- The Breakpoint i18n fallback test was fixed in #2025 (updated expectations to the now-translated Spanish strings, plus a new missing-locale-file fallback test).
- The `pnpm audit` failure (browserslist/jsondiffpatch CVEs) was fixed in #2035, via a scoped `ai>jsondiffpatch` override and `minimumReleaseAgeExclude` entries.

Both landed on `main` before this branch did, and `main`'s CI is now green on both fronts. This branch is now conflicting with `main` and merging it would risk regressing those fixes, so closing rather than rebasing.

PR #2045, opened 2026-09-04
docs: Standardize confidential key derivation on the main wallet

gitteri:
## Summary

Updates the confidential-transfer docs to the wallet-only key derivation standard, now matching the enforced no-seed SDK APIs: one wallet maps to one ElGamal keypair and one AES key across all mints and token accounts, and the standard functions take no seed at all so clients cannot accidentally derive different keys.

- JS snippets use the new sign-once deriveConfidentialKeys({ signer }) from @solana-program/token-2022/confidential (one signature over the constant solana-conf-bal/v1 message derives both keys). The local WASM-wrapping helper in the examples is renamed deriveConfidentialWasmKeys so it doesn't shadow the import. Tracks solana-program/token-2022#1432.
- Rust snippets use the no-seed derive_confidential_keys(&signer). Tracks solana-program/zk-elgamal-proof#533.
- The integration guide's key-management section explains the no-seed standard, keeps the note that sharing wallet-bound keys discloses all of an owner's confidential balances, and adds the wallet-side guidance: expose derivation through a dedicated flow and refuse generic signMessage requests for messages starting with solana-conf-bal/v1.
- Also fixes the integration guide importing the derivation helpers from the package root: they only exist on the /confidential subpath.

Merge ordering: this should land after the SDK PRs above ship in releases (@solana-program/token-2022 with deriveConfidentialKeys, solana-zk-sdk with the no-seed derive_confidential_keys), so the docs never reference unpublished APIs.

Localized mirrors (pl, vi, ja, etc.) are intentionally untouched: they flow through the lingo i18n pipeline.

Refs: ECO-504

### DIFF
--- a/apps/docs/content/docs/en/tokens/extensions/confidential-transfer/apply-pending-balance.mdx
+++ b/apps/docs/content/docs/en/tokens/extensions/confidential-transfer/apply-pending-balance.mdx
@@ -153,7 +153,7 @@ fn main() -> Result<()> {
     let (_mint, token_account) = setup_pending_balance_account(&rpc_client, &owner, amount, decimals)?;
 
     // Derive the owner's keys to decrypt the current balances.
-    let (elgamal_keypair, aes_key) = derive_confidential_keys(&owner, &token_account.to_bytes())
+    let (elgamal_keypair, aes_key) = derive_confidential_keys(&owner, b"")
         .map_err(|e| anyhow::anyhow!("derive confidential keys: {e}"))?;
 
     let account_data = rpc_client.get_account(&token_account)?;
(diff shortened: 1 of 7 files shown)


gitteri:
Review findings addressed. The auditor examples derive from a separate auditor wallet (create-mint loads it from the auditor's keypair file so the secret stays recoverable; the operation pages use a throwaway with a one-line note pointing at create-mint). The legacy-seeds section, the per-page warning callouts and the ForOwnerMint bullet are removed rather than expanded: nothing is deployed on the old derivations, so the docs describe the standard path only.

catmcgee:
the snippets now use the new derivation APIs, but the dependency blocks still select older SDK releases and exclude the new versions, can we update the Rust and JS dependencies in all six examples when those releases land, otherwise copying the examples will still fail after this merges

catmcgee (approved):

PR #2097, opened 2026-09-12
[DO NOT MERGE] chore(content): mark the larger transaction sizes upgrade live

amilz:
Stack 4/4, on top of #2096.

- Moves the upgrade page to `stage: live` and rewrites the activation framing, metrics label, and operator guidance now that the `txv1` gate has activated.

Note: the page hardcodes Sept 15 @ 01:20 UTC for epoch 1035 — needs a touch-up if that slips. Hold until v1 activates on mainnet.


amilz:
cancelled in lieu of #2107

PR #2126, opened 2026-09-17
Add sBPFv3 Upgrades Page

datasalaryman:


jacobcreech:
Can we update this to have language geared to a broader audience?

Woody4618:
Nice thanks. 

image

4.4 should be above 4.3 i think. 

Could you formulate this in a way that highlights the benefit of the change a bit more?

Something like:

“Brings Solana closer to modern/upstream eBPF conventions”
“Moves Solana toward a simpler, more maintainable runtime by standardizing future program deployments on SBPFv3 and newer.”

Woody4618 (approved):

PR #1816, opened 2026-07-31
Add Solana Scholars program page and application form

wattenhoferSOL:
### Problem

Solana Scholars — a funded 3-month research-internship program for PhD
students, run in collaboration with Solana engineers and researchers — had
no presence on solana.com. There was nowhere to point interested candidates
and no way to collect applications.

### Summary of Changes

Adds two routes to `apps/web`:

- `/scholars` — program information page
- `/scholars/apply` — application form

Both are self-contained (React/Next + one scoped CSS module, no
shared-package imports, no new env variables). The form posts applications
to the program's private Google Sheet via a Google Apps Script web app.
Security-relevant notes: the Apps Script endpoint URL appears in client
source by design — it is a write-only intake endpoint, not a credential;
the sheet itself stays private, and the endpoint can be rotated by
redeploying the script. Submissions are validated client-side and again
in the script (required fields, length caps).

---

### Change classification (SDLC §2)

- [ ] **Critical**
- [x] **Standard** — production-facing, non-critical (new public routes
      with a form)
- [ ] **Low**

### Security checklist

- [x] No secrets, tokens, or credentials in source, env files, or CI logs.
      (The Apps Script URL in `ApplyForm.tsx` is a public intake endpoint,
      not a credential.)
- [x] Input from users or external services is validated before use; no
      `dangerouslySetInnerHTML` / unsanitized HTML on untrusted input.
      (Native constraint validation client-side; the Apps Script re-checks
      required fields and caps field length server-side.)
- [x] New or updated dependencies are justified below, and `pnpm audit`
      passes (no new High/Critical advisories).
- [x] Errors are handled explicitly — failed submissions surface an error
      state in the form; the intake script returns `{ok:false}` on invalid
      input.
- [ ] For **Critical** changes: n/a

none

n/a

PR #2151, opened 2026-09-21
Polish Solana Scholars pages

sudo-dequoter:
## Summary
- refine the Scholars program and application page layouts
- add the PixelBlast hero treatment with a resilient WebGL fallback
- add section entrance motion with reduced-motion support

## Validation
- `pnpm --filter solana-com test`
- `pnpm --filter solana-com check-types`
- `pnpm --filter solana-com lint`
- `pnpm --filter solana-com build`

## Context
Stacked on #1816. Merging this PR into `scholars-page` updates the existing Scholars PR without duplicating its base changes.


rogerANZA:
Thanks for your work. Looks good to me, but I only tested a local copy. In particular I cannot test whether an actual application lands in some useful HR tool that we can use to screen (and possibly reject) the candidates.

One thing I would change: The middle part of the main page feels a bit repetitive. I would probably remove the whole part "How it works" (01-04). This is already somewhere else. Shorter is better in my opinion.

jacobcreech (approved):

